PR #16, opened 2016-07-03
update html elements

wormen:


icebob:
What is this PR? Could you make a sample fiddle with this feature?

wormen:
there is added to the developer directory

Issue #31, opened 2016-08-10
Add independent vanilla field components


icebob:
Create fields with vanilla libs which have no jQuery and Bootstrap dependencies.

- [X] Select field -> [vue-multiselect](https://github.com/monterail/vue-multiselect)
- [x] DateTime field -> [Pikaday](https://github.com/dbushell/Pikaday)
- [x] Masked input field -> [VanillaMasker](https://github.com/BankFacil/vanilla-masker) or [cleave.js](https://github.com/nosir/cleave.js)
- [x] Slider field -> [noUiSlider](https://github.com/leongersen/noUiSlider)

For these developments I created an `independent` branch. Every PR needs to add to this branch.

https://github.com/icebob/vue-form-generator/tree/independent

lionel-bijaoui:
noUiSlider is almost ready.
By default, noUiSlider return a float so I need a way to check if `validators.integer` is activated so I can `parseInt` the value if necessary.

lionel-bijaoui:
I will go with cleave.js, the API is much easier to work with and it is way more popular.

icebob:
Great!

icebob:
@lionel-bijaoui in the next 7 days I will on holiday, so my response will be longer, but I will check your commits as soon as possible :+1:

lionel-bijaoui:
That's perfect, I'm going on holiday too :)
I hope to do a PR for cleave and pikaday before the end of the day (just to initiate your feedback on them, I need some help).
Have a nice holiday !

icebob:
@lionel-bijaoui Thanks! You have a nice holiday too!

icebob:
@lionel-bijaoui I merged your fields. Fix lint errors and fix issues what I commented in your PR. So please rebase your fork.

lionel-bijaoui:
Thank you, I was about to start working on this. Hope you had a great
holiday.

lionel-bijaoui:
@icebob should I destroy my branches ?

icebob:
Yes, if you want.

icebob:
I will merge this `independent` branch to master

lionel-bijaoui:
So soon ? Don't we need the unit test and all ?

icebob:
This merge doesn't mean it will be a release :)
And need some tests for other fields too #40

icebob:
Merged with #42

PR #38, opened 2016-08-12
:new: new: new field type: Cleave. Reorganization of dev index file.

lionel-bijaoui:
#31 
This is not yet fully functional. Cleave.js use [functions](https://github.com/nosir/cleave.js/blob/master/doc/public-methods.md#getrawvalue) to set/get value. I was not able to save the data into the model (I tried different solution that you can see commented). I don't know if the dev example need an update. Please can you take a look ?
Also, I have modified the organization of the script files in index for a better clarity.

### DIFF
--- a/src/fields/fieldCleave.vue
+++ b/src/fields/fieldCleave.vue
@@ -1,0 +1,103 @@
+<template lang="jade">
+	input.form-control(type="text")
+</template>
+
+<script>
+import abstractField from "./abstractField";
+import { defaults } from "lodash";
+
+export default {
+	mixins: [abstractField],
+
+	data() {
+		return {
+			cleave: null
+		};
+	},
+
+	watch: {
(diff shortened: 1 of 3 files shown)


icebob:
I think it is very simple. Just add `v-model="value"` to the input. I tried and it is working. So you can remove the watch and methods comment blocks.

```jade
<template lang="jade">
	input.form-control(type="text", v-model="value")
</template>
```

src/fields/fieldCleave.vue
```diff
@@ -0,0 +1,103 @@
+<template lang="jade">
+	input.form-control(type="text")
```
icebob: Please add the general attributes too:
```jade
input(...,  :readonly="schema.readonly", :disabled="disabled", :placeholder="schema.placeholder")
```

PR #39, opened 2016-08-12
new field type: Pikaday

lionel-bijaoui:
Same as #38, it is not yet fully functional. You can change the option and it will appear but I can't link it to the model and/or value. 

#31

### DIFF
--- a/src/fields/fieldCleave.vue
+++ b/src/fields/fieldCleave.vue
@@ -1,0 +1,103 @@
+<template lang="jade">
+	input.form-control(type="text")
+</template>
+
+<script>
+import abstractField from "./abstractField";
+import { defaults } from "lodash";
+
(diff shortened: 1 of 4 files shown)


src/fields/fieldCleave.vue
```diff
@@ -0,0 +1,103 @@
+<template lang="jade">
+	input.form-control(type="text")
```
icebob: Please add the general attributes too:
```jade
input(...,  :readonly="schema.readonly", :disabled="disabled", :placeholder="schema.placeholder")
```

icebob:
The solution what I wrote to cleave, it is working too here.
```jade
    input.form-control(type="text", v-model="value")
```

icebob:
Merged.

Issue #40, opened 2016-08-21
Create unit tests for new fields


icebob:
- [x] fieldCleave
- [x] fieldMasked
- [x] fieldNoUiSlider
- [x] fieldPikaday
- [x] fieldSlider
- [x] fieldSpektrum
- [x] fieldVueMultiSelect
- [x] fieldGoogleAddress

lionel-bijaoui:
I'm going to try after I read a bit more about Karma

icebob:
OK, but firstly I have to solve the external lib loading problem.

icebob:
I added external libs to karma. So now can be improve new field unit tests. https://github.com/icebob/vue-form-generator/commit/776f6c479fec8105df55d69288a310cf58b2386c

lionel-bijaoui:
(I was about to suggest that :smile: )

---
> Passing arrow functions (“lambdas”) to Mocha is discouraged

From the [doc](http://mochajs.org/)
Should we change that ?

icebob:
Oooo.... what?!?!

lionel-bijaoui:
basically, don't do this:
```
describe('my suite', () => {
  it('my test', () => {
    // should set the timeout of this test to 1000 ms; instead will fail
    this.timeout(1000);
    assert.ok(true);
  });
});
```
but this:
```
describe('my suite', function() {
  it('my test', function() {
    this.timeout(1000);
    assert.ok(true);
  });
});
```

icebob:
You can use arrows because we use webpack as preprocessor and webpack use babel

lionel-bijaoui:
Unit test are going through Webpack ? Wow ok, I didn't know... Webpack does everything :astonished:

icebob:
yes :) https://github.com/icebob/vue-form-generator/blob/master/test/unit/karma.conf.js#L35

lionel-bijaoui:
Sorry, I missed that totally

icebob:
Nop, so you can use any ES6 feature

lionel-bijaoui:
Sweet!

icebob:
what field test are you make? Because I plan to make an other field tests

lionel-bijaoui:
I'm starting with #36

icebob:
**What we need to test in this external fields:**

- [ ] check HTML elements (example: https://github.com/icebob/vue-form-generator/blob/master/test/unit/specs/fields/fieldSelect.spec.js#L48 )

[...]

PR #44, opened 2016-08-23
noUiSlider unit test

lionel-bijaoui:
:construction: WIP
I'm able to wait for the initialization of noUiSlider thanks to [these slides](http://www.slideshare.net/coulix/vuejs-testing) (n°9 was key).
By appending the `vm` to the document, it trigger `ready`, which initialize noUiSlider.
Then, I use `vm.$nextTick()` to make sure noUiSlider had the time to do it's business (transforming a simple div into a nice slider), ready to test things !

For now, nothing important is tested. This was just to show you this solution to test with external libs.


icebob:
Thanks. Good solution.

Btw, what we test in this external fields: https://github.com/icebob/vue-form-generator/issues/40#issuecomment-241689467

test/unit/specs/fields/fieldNoUiSlider.spec.js
```diff
@@ -38,5 +38,25 @@ describe("fieldNoUiSlider.vue", () => {
...
+				it("should contain a complex div element", (done) => {
+					vm.$nextTick( () => {						
+						expect(input).to.be.defined;
+						expect(input.classList.contains("noui-target")).to.be.true;	
+						done();
+					});
+				});
+
+			} else {
+				this.skip();
```
icebob: Travis failed. Missing done()

icebob:
Your name doesn't visible in contributors page. https://github.com/icebob/vue-form-generator/graphs/contributors

> Contributions to master, excluding merge commits

I think maybe your commits are in a separated branches instead of master of your fork. What do you think?

lionel-bijaoui:
I've skipped a lot of test, noUiSlider is a bit weird...

icebob:
Ok, but I write some comment to code, ok?

lionel-bijaoui:
I'm sorry, what ?

test/unit/specs/fields/fieldNoUiSlider.spec.js
```diff
@@ -36,7 +36,71 @@ describe("fieldNoUiSlider.vue", () => {
...
+			if (window.noUiSlider) {
+				vm.$nextTick( () => {
+					let handle = input.querySelectorAll(".noUi-handle");
+					expect(handle.length).to.be.equal(1);					
+					expect(handle[0]).to.be.defined;
+					// expect(input.classList.contains("noui-target")).to.be.true;
+					done();
+				});
+			} else {
+				this.skip();
```
icebob: I think here, we throw an exception if noUISlider is not loaded instead of skip

test/unit/specs/fields/fieldNoUiSlider.spec.js
```diff
@@ -36,7 +36,71 @@ describe("fieldNoUiSlider.vue", () => {
...
+		it.skip("should contain the value", (done) => {
+			vm.$nextTick( () => {
+				let origin = input.querySelectorAll(".noUi-origin")[0];				
+				expect(origin.style.left).to.be.within("70%", "90%");
+				done();
+			});
+		});
+
+		before( () => {
+			vm.model = { rating: 10 };
```
icebob: This modifications place in `it` instead

test/unit/specs/fields/fieldNoUiSlider.spec.js
```diff
@@ -36,7 +36,71 @@ describe("fieldNoUiSlider.vue", () => {
...
+		});
+
+		it("handle value should be the model value after changed", (done) => {
+			vm.$nextTick( () => {
+				let origin = input.querySelectorAll(".noUi-origin")[0];
```
icebob: Please use `querySelector` if you need the first element. Because `querySelectorArr()[0]` throw exception if the result is empty. However `querySelector` only return with null.

test/unit/specs/fields/fieldNoUiSlider.spec.js
lionel-bijaoui: If I do that, the value don't change

test/unit/specs/fields/fieldNoUiSlider.spec.js
lionel-bijaoui: Should I do something like this ?
```
if (typeof(window.noUiSlider) === 'object')
```

test/unit/specs/fields/fieldNoUiSlider.spec.js
icebob: Hmm, strange. I will test it.
But `before()` runs before all `it`, so this solution is not precise.

test/unit/specs/fields/fieldNoUiSlider.spec.js
icebob: No, the check is good, but change the `this.skip()` to `throw new Exception("Library is not loaded");`

PR #47, opened 2016-08-24
small fix

lionel-bijaoui:
Very small stuff, should be quickly merge if you agree with the changes

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,7 +17,7 @@
     "e2e": "node test/e2e/runner.js",
     "pretest": "npm run lint",
     "test": "npm run unit",
-    "ci": "karma start test/unit/karma.conf.js --auto-watch --no-single-run --reporters=dots"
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -17,7 +17,7 @@
     "e2e": "node test/e2e/runner.js",
     "pretest": "npm run lint",
     "test": "npm run unit",
-    "ci": "karma start test/unit/karma.conf.js --auto-watch --no-single-run --reporters=dots"
+    "ci": "karma start test/unit/karma.conf.js --auto-watch --no-single-run --reporters=dots --port=9877"
```
icebob: Why need it?

lionel-bijaoui: As I said in the commit, I allow to use both `npm run test` and `npm run ci` at the same time without warning. It is a very small stuff, but since I use them all the time, I want to avoid any interference even if very unlikely.

icebob:
ok, merged

PR #48, opened 2016-08-25
Fix for VueMultiselect

lionel-bijaoui:
While doing the documentation, I noticed some missing options for VueMultiselect.
I also removed `min` from dev schema because it was not used. You can only prevent empty (with `allowEmpty: false`), but not specify a minimum number of option to select, only a maximum.


icebob:
OK, merged

PR #52, opened 2016-08-26
fix: fragment instance warning

lionel-bijaoui:
Fix for #45 
It seem like a `fieldset` don't count as a root element for Vue. I added a `div` and it fix the warning.

### DIFF
--- a/src/formGenerator.vue
+++ b/src/formGenerator.vue
@@ -1,4 +1,5 @@
 <template lang="jade">
+div
 	fieldset.vue-form-generator(v-if="schema != null")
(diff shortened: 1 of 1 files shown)


icebob:
Did you test it with a modal?

lionel-bijaoui:
No, but Vue.js devtools show a warning when a component is a fragment instance. This simple change fix that warning. I assume it will fix a lot of problem. Fragment instance is a pain to work with in Vue.js.
I hope @blackfyre will test this and give his feedback.

icebob:
Ok, just I test it before merge.

icebob:
Merged

lionel-bijaoui:
@icebob Did you do an build then commit since this PR ?

Issue #25, opened 2016-08-08
update docs for v0.4.0


icebob:
- [x] vue-multiselect field #20
- [x] fieldSubmit #26 
- [x] customizable styles #29 
- [x] fieldCleave #31
- [x] fieldPikaday #31
- [x] fieldNoUiSlider #31
- [x] fieldGoogleAddress #33 
- [x] visible & disabled, values schema functions description
- [x] ES5 examples (change `visible(model) {...}` to visible: `function(model) {...}` )

lionel-bijaoui:
Need help with the doc ?

icebob:
Maybe :smiley:

lionel-bijaoui:
Go on, tell me: how can I help?

icebob:
The docs on gitbook.com. I don't know it enough good, I don't know the contribution is available.

lionel-bijaoui:
I just created an account on gitbook, please authorize me so I can try.

icebob:
OK, I added.

icebob:
Please check, can you modify the docs?

lionel-bijaoui:
YES ! :tada:

icebob:
Great!

lionel-bijaoui:
Maybe we should specify "vanilla vs jQuery" in the `installation.md`? I fear the dependency list could be misunderstood and seem daunting. Maybe dependency should be specified only for each field ?

icebob:
Yes we need specify dependencies for every each field

lionel-bijaoui:
I'm going to create a branch on gitbook for the v0.4.0 and start modifying the doc

icebob:
Great! Thanks!

lionel-bijaoui:
I don't understand what `multiple` do, can you explain please ?
It is listed in `component.md` and `fields/field_properties.md`, is it two different options, or a mistake ?

icebob:
You can test in in dev mode. Use CTRL key and select multiple rows in table. In this case you can edit fields for multiple models. And component shows fields where `multiple: true`.

But currently it is not complete, because validator can't handle it correctly yet.

lionel-bijaoui:
So the main `multi` activate the option, and field level `multi` activate each field to be part of it ?

icebob:
Has `multi` and `multiple`. Both exists.
Field has property `multi`.
Component has `multiple`˛which activate this "multi mode".
See in dev example:
```js
[...]

PR #54, opened 2016-08-29
fix: buttonText instead of caption

lionel-bijaoui:
#25 
I changed dev schema and fieldSubmit


icebob:
Please fix fieldSubmit.spec.js too. It contains `caption`

icebob:
Thanks. Merged

Issue #59, opened 2016-08-31
Pikaday field: Invalid date


lionel-bijaoui:
When I use a pikaday field with no special options, I get "Invalid date" in the field.
I can see the good value for a fraction of a second but then it get replaced by this message.
This problem don't appear with the dev environment, and I suspect it has something to do with Moment.js (or the lack of it to be precise).
I experimented a bit with `format`, but nothing change that.
For now I will send a PR which don't add any default values (except `field`) in pikadayOptions.

icebob:
Could you make a fiddle with this issue?

lionel-bijaoui:
> [Fiddle](https://jsfiddle.net/0mg1v81e/38/)

Found more things:
- if the model is not defined :arrow_right: empty field
- if the model is defined with an empty `String` :arrow_right: "Invalid date" appear in the field
- if the model is defined with a `String` in the same format defined :arrow_right: the date appear into the field but at first (without any interaction) the format is not respected (eg. "Thu Sep 01 2016" instead of "2016-09-01"). When a value is selected, for a split second the non formatted value appear then the good format replace it.
- if the model is defined with an empty `Object` :arrow_right: the date appear into the field but at first (without any interaction) the format is not respected (eg. "Thu Sep 01 2016" instead of "2016-09-01"). When a value is selected, for a split second the non formatted value appear then the good format replace it.

So, even when the format is good or with an empty `Object`, the first value is not respecting the desired format, and this undesired format still appear for a split second.
We might need to capture the click events and stop propagation or use `onSelect` callback to launch `formatValueToField`. Also maybe launch `picker.setDate('2015-01-01')` (or `picker.setMoment(moment('14th February 2014', 'DDo MMMM YYYY'))`) at the ready hook just after initialization.
What do you think ?

icebob:
They say: 
> No dependencies (but plays well with Moment.js)

lionel-bijaoui:
I know, but try to remove moment.js from the dev test files and use Pikaday to select a date. You will get "Invalid date". This may not be our fault, because I don't see where we use it without importing it first.
What do you think we should do ? Change the documentation ? Try to find and correct the bug ?

icebob:
For quick solution please add the momentjs to pikaday as dependencies in the documentation. If I have more time, I will try to find the exact problem.

lionel-bijaoui:
I edited the doc and pointed to this issue

icebob:
[...]

PR #60, opened 2016-08-31
:bug: bug: "Invalid date" message in pikaday field.

lionel-bijaoui:
Change dev schema for a simple one. Delete default in pikaday field.
#59

### DIFF
--- a/dev/schema.js
+++ b/dev/schema.js
@@ -303,48 +303,44 @@ module.exports = {
 			model: "dob",
 			required: true,
 			placeholder: "User's birth of date",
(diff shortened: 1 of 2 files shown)


icebob:
Why solve it the problem?

icebob:
I closed because it is not solved the problem.

lionel-bijaoui:
I agree it does not solve the problem.
But can you remove line 52 to 84 of `fieldPikaday.vue` ? There is no need for this now that the doc is ready. We don't want to override the default of the plugin with the same option as the default of the plugin, it is pointless.

icebob:
Merged

PR #63, opened 2016-09-01
enhancement: add missing "autocomplete", "placeholder" and "readonly"…

lionel-bijaoui:
Add missing "autocomplete", "placeholder" and "readonly" attr to input fields. Ordered alphabetically.

### DIFF
--- a/src/fields/fieldNumber.vue
+++ b/src/fields/fieldNumber.vue
@@ -1,5 +1,5 @@
 <template lang="jade">
-	input.form-control(type="number", v-model="value", number, :autocomplete="schema.autocomplete", :disabled="disabled", :max="schema.max", :min="schema.min", :readonly="schema.readonly")
+	input.form-control(type="number", v-model="value", number, :autocomplete="schema.autocomplete", :disabled="disabled", :max="schema.max", :min="schema.min", :placeholder="schema.placeholder", :readonly="schema.readonly")
 </template>
 
 <script>
--- a/src/fields/fieldNumber.vue
+++ b/src/fields/fieldNumber.vue
@@ -1,5 +1,5 @@
 <template lang="jade">
-	input.form-control(type="number", v-model="value", number, :min="schema.min", :max="schema.max", :readonly="schema.readonly", :disabled="disabled", :placeholder="schema.placeholder")
+	input.form-control(type="number", v-model="value", number, :autocomplete="schema.autocomplete", :disabled="disabled", :max="schema.max", :min="schema.min", :placeholder="schema.placeholder", :readonly="schema.readonly")
 </template>
 
 <script>
--- a/src/fields/fieldNumber.vue
+++ b/src/fields/fieldNumber.vue
@@ -1,5 +1,5 @@
 <template lang="jade">
-	input.form-control(type="number", v-model="value", number, :autocomplete="schema.autocomplete", :disabled="disabled", :max="schema.max", :min="schema.min", :placeholder="schema.placeholder", :readonly="schema.readonly")
+	input.form-control(type="number", v-model="value", number, :autocomplete="schema.autocomplete", :disabled="disabled", :max="schema.max", :min="schema.min", :readonly="schema.readonly")
 </template>
 
 <script>
--- a/src/fields/fieldPassword.vue
+++ b/src/fields/fieldPassword.vue
@@ -1,5 +1,5 @@
 <template lang="jade">
-	input.form-control(type="password", v-model="value", :autocomplete="schema.autocomplete", :disabled="disabled", :readonly="schema.readonly")
+	input.form-control(type="password", v-model="value", :autocomplete="schema.autocomplete", :disabled="disabled", :placeholder="schema.placeholder", :readonly="schema.readonly")
 </template>
 
 <script>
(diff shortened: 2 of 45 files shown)


src/fields/fieldNumber.vue
```diff
@@ -1,5 +1,5 @@
 <template lang="jade">
-	input.form-control(type="number", v-model="value", number, :min="schema.min", :max="schema.max", :readonly="schema.readonly", :disabled="disabled", :placeholder="schema.placeholder")
+	input.form-control(type="number", v-model="value", number, :autocomplete="schema.autocomplete", :disabled="disabled", :max="schema.max", :min="schema.min", :readonly="schema.readonly")
 </template>
```
icebob: Why removed the placeholder from here?

src/fields/fieldPassword.vue
```diff
@@ -1,5 +1,5 @@
 <template lang="jade">
-	input.form-control(type="password", v-model="value", :readonly="schema.readonly", :disabled="disabled", :placeholder="schema.placeholder")
+	input.form-control(type="password", v-model="value", :autocomplete="schema.autocomplete", :disabled="disabled", :readonly="schema.readonly")
 </template>
```
icebob: Why removed the placeholder from here?

lionel-bijaoui: Sorry the [french documentation](https://developer.mozilla.org/fr/docs/Web/HTML/Element/Input) was obsolete and wasn't listing `password` and `number` as compatible with `placeholder`. I will fix that.

icebob:
Please add tests to the new attributes like `disabled` attr.

lionel-bijaoui:
I had a little bit of hard time, but I was able to simplify the test for attributes to this:
```js
import { createVueField, trigger, /*new thing in util.js --> */ attributesList } from "../util";

describe.only("check optional attribute", () => {
            // here you name which attribute you want to test and that's it.
            let attributes = ["autocomplete", "disabled"];

            attributes.forEach(function(name) {
                it('should set ' + name, function(done) {
                    let schematic;
                    let attr = attributesList[name];

                    if (attr.field) {
                        schematic = field;
                    } else {
                        schematic = schema;
                    }
                    schematic[name] = attr.before;
                    vm.$nextTick(() => {
                        expect(input[name]).to.be.equal(schematic[name]);
                        schematic[name] = attr.after;
                        done();
                    });
                });
            });
        });
```
```js
// util.js
export let attributesList = {
    "autocomplete": { before: "on", after: "off" },
    "disabled": { before: true, after: false, field: true },
    "placeholder": { before: "Field placeholder", after: "" },
    "readonly": { before: true, after: false }
    // etc...
};
```

lionel-bijaoui:
My tests fail if the properties are not set in the `schema`

lionel-bijaoui:
Woops my bad

icebob:
> My tests fail if the properties are not set in the schema
Please explain, I don't understand.

lionel-bijaoui:
describe("check template", () => {
        let schema = {
            type: "masked",
            label: "Phone",
            model: "phone",
            // autocomplete: "off", // <-- not set here
            placeholder: "",
            readonly: false,
            cleaveOptions: {
                phone: true,
                phoneRegionCode: "HU",
            }
        };
        let model = { phone: "30 123 4567" };
        let input;

        before(() => {
            createField(this, schema, model, false);
            input = el.getElementsByTagName("input")[0];
        });

        it('should set autocomplete', function(done) {
            schema.autocomplete = "on";
            vm.$nextTick(() => {
                expect(input.autocomplete).to.be.equal(schema.autocomplete);  // <-- fail here
                schema.autocomplete = "off";
                done();
            });
        });
    });

icebob:
Yes, because first case the "autocomplete" property is not reactive. Use schema.$set("autocomplete", "on") or field.schema.$set("autocomplete", "on")

lionel-bijaoui:
I tried it and it trigger other errors. I want to finish it so I'm going to do it by initializing for now.

test/unit/specs/fields/fieldCleave.spec.js
```diff
@@ -41,9 +42,9 @@ describe("fieldCleave.vue", function() {
 			expect(input).to.be.defined;
 			expect(input.type).to.be.equal("text");
 			expect(input.classList.contains("form-control")).to.be.true;
-			expect(input.placeholder).to.be.equal(schema.placeholder);	
-			expect(input.readOnly).to.be.false;	
-			expect(input.disabled).to.be.false;	
+			// expect(input.placeholder).to.be.equal(schema.placeholder);
+			// expect(input.readOnly).to.be.false;
```
icebob: Please remove this old commented rows (other files too).

lionel-bijaoui: Yeah, I wasn't sure you wanted to delete them or not, so meanwhile, I commented them. I will clean things now !

icebob: You are mindful. But if we covered with your new code we can delete them. Because after 3 months, we won't know it is a skipped tests (what need to fix) or unneccessary :)

lionel-bijaoui: Ok, make sense, sorry about that.

icebob: Nop, if you ready, I will merge it. :tada:

icebob:
So can be merge? (after green travis)

Issue #50, opened 2016-08-25
create native HTML5 fields


icebob:
Because currently there is no vanilla datetime picker. Pickaday is only date selector.

- [x] create new field common type which has a "inputType" property. (break changes)
- [x] rename `datetime` field to `datetimePicker` (break changes)
- [x] rename `slider` field to `rangeSlider` (break changes)
- [x] update docs (add new radio values description)

Available types
- [x] date
- [x] datetime
- [x] datetime-local
- [x] time
- [x] month
- [x] week

Other HTML5 types
- [x] url
- [x] tel
- [x] radio

Set this inputType to the field type attributes:
```jade
input( :type="schema.inputType")
```

lionel-bijaoui:
Same thing should be done with slider field. It should be renamed "ion.rangeSlider" or "rangeSlider".
Also a "radio" field ! Switch is cool, but sometime you need more the 2 values.
And "tel" for a native masked/cleave for phone number only.
Last one is "url".
I would also like to point out that we need more HTML5 attributes:
- autofocus
- inputmode
- selectionDirection
- spellcheck

specific to text, search, tel, url or email:
- pattern
- size (also for password)

:boom: *More HTML5!* :boom:

icebob:
HTML5 range is exist as `fieldRange`. I added the other types to the tasklist.

what is the `size` attr?

lionel-bijaoui:
Yeah, sorry I forgot to remove that sentence.

> **size**
The initial size of the control. This value is in pixels unless the value of the type attribute is text or password, in which case, it is an integer number of characters. Starting in HTML5, this attribute applies only when the type attribute is set to text, search, tel, url, email, or password; otherwise it is ignored. In addition, the size must be greater than zero. If you don't specify a size, a default value of 20 is used. HTML5 simply states "the user agent should ensure that at least that many characters are visible", but different characters can have different widths in certains fonts; in some browsers, a certain string with x characters will not be entirely visible even if size is defined to at least x.

[Source](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input)

I think the goal for v1.0.0 is to have all attributes exposed for native HTML5 fields.

icebob:
I think, the source is not neccessary, because the default `width` is 100%. So it is overwrite the `size` attr.
And you can change this width via `styleClasses` property in schema

lionel-bijaoui:
[...]

PR #70, opened 2016-09-14
new: new field 'input'

lionel-bijaoui:
#50 this is my proposal for this fields.
The test are not amazing for now (they accept any attribute regardless of the input type).

### DIFF
--- a/src/fields/fieldInput.vue
+++ b/src/fields/fieldInput.vue
@@ -37,7 +37,35 @@
 	import abstractField from "./abstractField";
 
 	export default {
-		mixins: [ abstractField ]
+		mixins: [ abstractField ],
+		methods: {
+			formatValueToField(value) {
+				switch(this.schema.inputType){
+					case "date":
+						return moment(value).format("YYYY-MM-DD");
+					case "datetime":
+						return moment(value).format();
+					case "datetime-local":
+						return moment(value).format("YYYY-MM-DDTHH:mm:ss");
+					default:
+						return value;
+				}
+			},
+			formatValueToModel(value) {
+				console.log(this.schema.inputType, typeof value);
+				if (value != null) {
+				    if (this.schema.inputType === "date" ||
+				        this.schema.inputType === "datetime" ||
+				        this.schema.inputType === "datetimelocal") {
+						return new Date(value).getTime();
+				    }else{
+				    	return value;
+				    }
+				} else {
+					return value;
+				}
+			}
+		}
 	};
 	
 </script>
--- a/src/fields/fieldInput.vue
+++ b/src/fields/fieldInput.vue
@@ -34,6 +34,7 @@
 </template>
 
 <script>
+/* global moment */
 	import abstractField from "./abstractField";
 
 	export default {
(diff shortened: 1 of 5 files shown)


icebob:
Thank you for your PR. I need more time to check it, ok?

icebob:
I think, we can't handle all types with one field. Because date* types there are extra properties and methods to format date to the input and to the model like [here](https://github.com/icebob/vue-form-generator/blob/master/src/fields/fieldDateTime.vue#L28)

Do you have any idea how to solve it?

lionel-bijaoui:
A simple check to see the this.schema.format could be enough:
```js
if (typeof this.schema.format !== "undefined" && value != null) {
	return moment(value, this.schema.format).format(this.getDateFormat());
}
return value;
```
If the user don't want to format the date or if there is no reason to use format (for non date like input), it simply return the raw value.

icebob:
Not enough, because current date fields (datePicker, pickaday) are stored the data in Unix format without `format`. If you set `format` property it will store as a formatted date `string`. But HTML5 date* field raw format is string. So we always should format this raw values.

I created a fiddle to check the raw values of HTML5 date fields: https://jsfiddle.net/ndz05ku4/2/

lionel-bijaoui:
What about using `inputType` ?
```js
if (value != null) {
    if (this.schema.inputType === date ||
        this.schema.inputType === datetime ||
        this.schema.inputType === datetimelocal ||
        this.schema.inputType === time ||
        this.schema.inputType === week ||
        this.schema.inputType === month) {
            if (typeof this.schema.format === "undefined"){
                // Unix formating
            }else{
                return moment(value, this.schema.format).format(this.getDateFormat());
            }
    }
}
return value;
```

icebob:
Ok, but skip `month` and `week`. They are not full date, can't be format with moment. Stays in string.

lionel-bijaoui:
I'm skipping `time` too

lionel-bijaoui:
Progress: `date`, `datetime` and `datetime-local` can use the same model as Pikaday, dateTime (with date format) without problem. Modifying one or the other end in the same model format (UNIX) and update value.
`time` work well with dateTime (with time format), although Chrome don't allow to input seconds (but display them if set). Edge is fine with seconds.
`month` and `week` are treated as string.

icebob:
Great thank you!

lionel-bijaoui:
@icebob Can I add a new schema to the commit ?
I have reorganize the fields so that it is easier to use
[screenshot](https://cloud.githubusercontent.com/assets/3472596/18639210/3f4ba972-7e93-11e6-81f4-fb3c6a901dcb.png)

icebob:
Sure :)

lionel-bijaoui:
@icebob I think it is ready to merge, no ?

src/fields/fieldInput.vue
```diff
@@ -37,7 +37,35 @@
...
+					case "datetime":
+						return moment(value).format();
+					case "datetime-local":
+						return moment(value).format("YYYY-MM-DDTHH:mm:ss");
+					default:
+						return value;
+				}
+			},
+			formatValueToModel(value) {
+				console.log(this.schema.inputType, typeof value);
```
icebob: Please remove this log message and after it I will merge.

lionel-bijaoui: done

icebob: Thanks

lionel-bijaoui: You still need to merge, no ?

icebob: Just waited the green lights :)

lionel-bijaoui: sorry I did not get it :+1:

Issue #71, opened 2016-09-20
Unify value from noUiSlider


lionel-bijaoui:
There is a few problem with noUiSlider:
- [x] It can't dynamically add or remove an handle. It can't wait for the value from the model to know if this is a double handle or not. So we need an option to explicitly tell if this is double or not (like with rangeSlider). EDIT: done in #72
- [x] When using a single handle, noUiSlider format the data like this `[Number, null]`. It expect us to use `slider.noUiSlider.get()`. So, much like with date input, we need a `formatValueToField` and `formatValueToModel` to properly format data back and forth.

I'm on both issue and I will do a PR in the next days.

icebob:
Nice observation. Thank you

lionel-bijaoui:
Progress: I'm done with noUiSlider, I added some dynamic CSS class and CSS fix with scale.
There is something I can't do though: if you modify the same model with noUiSlider and rangeSlider, modification of rangeSlider make noUiSlider change too, but nit the contrary.
I tried different solutions ( almost lost my mind) and was not able to avoid infinite loops. I don't know why, and I give up, since there is no real reason to use both sliders.

lionel-bijaoui:
#73

PR #73, opened 2016-09-22
noUiSlider fix

lionel-bijaoui:
Fix some problem described in #71 and add some enhancement:
- to help style the field, some class are applied dynamically (for pips and tooltips)
- schema is updated along with data to generate `income` data and to uncomment rangeSlider (I accidentally commented it in a previous update)
- a small change on a condition to avoid "undefined value" warning

A unit test was deactivated since I was not able to trigger a change in model. :disappointed: 
I don't know how to simulate a click somewhere on the range slider or have it ready to use.
I give up for now, since the field clearly work as expected anyway. It represent too much work for me right now. Hope that's ok.

### DIFF
--- a/dev/data.js
+++ b/dev/data.js
@@ -45,6 +45,8 @@ module.exports = {
 			if (user.type == "business")
 				user.company = fakerator.entity.company();
 			
+			user.income = [ fakerator.random.number(50000), fakerator.random.number(50000, 100000)];			
+
 			res.push(user);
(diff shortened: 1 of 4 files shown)


icebob:
Thanks. But I create an issue with this skipped test, that we don't forget it.

lionel-bijaoui:
Thank you, I was about to create the issue ! :)

Issue #87, opened 2016-10-06
multiselect: put back custom-label & fix


icebob:
I [removed](https://github.com/icebob/vue-form-generator/commit/5dfc5bb866e8c713ab89be92c62a4645f7c765b4) `custom-label` property from multiselect because give errors.
If it props doesn't exist in schema, multiselect try to call and gives error.

lionel-bijaoui:
I had the same problem while working the v2, and found a solution. It need to at least (even when empty) return a function. I was returning `null` and it was not happy.
I will update my master and do a PR with the fix.

icebob:
Merged. Thanks!

PR #89, opened 2016-10-07
Toward vuejs v2.0

lionel-bijaoui:
🚧 This is WIP 🚧 
current problem by order of importance:
- Tests need to be rewritten in some way to fit vuejs 2.0 😅 
- selectEx: weird warning, problem is not yet identified.
- dev app: is not working well and can be simplified with ES6.
- validator: weird warning seem to indicate lack of information on initialisation. May be related to dev app or internal problem with vuejs v2.0. Some behavior have changed and things update faster which may or may not be the cause.
- noUiSlider: I'm updating it to v9.0 and there is breaking changes and cool new functionality (very useful for my work).
- vue multiselect should work now that #87 is implemented, but I need to make sure.
- updating plugins

Project board : https://github.com/icebob/vue-form-generator/projects/2


.babelrc
```diff
@@ -1,5 +1,5 @@
 {
-  "presets": ["es2015", "stage-0"],
+  "presets": ["es2015", "stage-2"],
   "plugins": ["transform-runtime"],
```
icebob: Why changed to stage-2 from 0?

src/fields/abstractField.js
```diff
@@ -1,93 +1,100 @@
...
+            },
+
+            set(newValue) {
+
+                if (isFunction(this.formatValueToModel))
+                    newValue = this.formatValueToModel(newValue);
+
+                if (isFunction(this.schema.set)) {
+                    this.schema.set(this.model, newValue);
+                    console.log("model-updated via schema", this.model[this.schema.model]);
```
icebob: Not guaranteed, that this.schema.model is defined. Need to check

webpack.dev.config.js
```diff
@@ -1,27 +1,21 @@
...
 	},
 	{
-		"test": /\.scss?$/,
-		"loader": "style!css!sass"
-	},
-	{
-		"test": /\.jade?$/,
-		"loader": "jade"
-	},
-	{
```
icebob: In Vue2 these loaders not neccessary?

test/unit/specs/VueFormGenerator.spec.js
```diff
@@ -10,24 +10,27 @@ let el, vm;
...
+	}).$mount(el);
 
 	// console.log(el);
 
 	return [el, vm];
 }
 
 describe("VueFormGenerator.vue", () => {
-
+	describe.only("nothing", () => {
```
icebob: It's misleading, because Travis will be green, but no correct testing. In development state there is no problem if Travis red, but show the real results and we know we need to fix it too. :wink:

.babelrc
lionel-bijaoui: stage-2 is the default in the vue-cli webpack preset. It is less bleeding edge, I guess ? I don't mind putting it back to were it was, I just did that when I was debugging to exclude differences between your config and the proposed webpack config.
You use ES7 features ?

src/fields/abstractField.js
lionel-bijaoui: BTW, what is this `schema.set` ? I might have missed it, but I don't recall any explanation about it in the doc.

test/unit/specs/VueFormGenerator.spec.js
lionel-bijaoui: Sorry about that, it was to test the build locally without changing too much. Normally, it is forbidden to push `.only()`. This will go away when I fix the test. I forget to put it back how it was.

webpack.dev.config.js
lionel-bijaoui: Well, we don't require any `.jade`, `.scss` or `.css` files. They were never necessary (vue-loader take care of this).
It is also part of my temporary debug changes.
It can came back we everything is working again.

.babelrc
icebob: Currently not, just I'm using stage-0 in my other projects too.

webpack.dev.config.js
icebob: Hmm, you are right. It's a habit :)

test/unit/specs/VueFormGenerator.spec.js
icebob: OK

icebob:
Hi, are you working on this upgrade yet?

lionel-bijaoui:
@icebob Sorry I was very busy at work. I will work on it soon.

icebob:
@lionel-bijaoui Do you have any new commit to this PR?
Because maybe next week I'll able to work on vue2 migration. And I want to merge this PR to the `next` branch.

lionel-bijaoui:
I'm sorry, I'm really squeezed tight at work. I'm not in my normal office and I'm not doing any dev work. It's kinda special for me for the next 2 months.
I'm going to take a look at my PR tomorrow. Worse case scenario, you can use my work as a base. Most of it work well. The unit tests are the biggest job along the validator system.
Sorry I can't be more helpful, I feel really bad about that 😞

icebob:
Thanks, no problem! Just upload your last state and I continue to migration. :wink:

lionel-bijaoui:
I updated vue and integrated all the new changes since the last push. The main problem is the validator that is validating too early. I don't know if it has access to the values or not.
With that, the test need to be fixed and updated.
Lastly, some libs were updated and need to be checked for changes (NoUiSlider for example can now accept an arbitrary number of handle + syntax changes)

icebob:
Thank you!

PR #103, opened 2017-01-12
Allow form fields required state to be controlled with a function

mpociot:
This PR allows you to use a callback function with model to control the `required` state of a form field.

Example to make a field only required, if another field is empty:

```js
{
    type: "text",
    label: "Phone mobile",
    model: "phone_mobile",
    required: function(model) {
    	return (model.phone_home === null);
    }
}
```

### DIFF
--- a/src/formGenerator.vue
+++ b/src/formGenerator.vue
@@ -108,7 +108,7 @@ div
 					disabled: this.fieldDisabled(field), 
 					readonly: field.readonly, 
 					featured: field.featured, 
-					required: field.required
+					required: this.fieldRequired(field)
 				};
 
 				if (isArray(field.styleClasses)) {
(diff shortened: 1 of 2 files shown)


icebob:
Good job! Thanks!

mpociot:
What do you think of having the ability to use callback functions for the other boolean checks as well? I would provide a callback for this too.

icebob:
Please tell which properties do you think?

mpociot:
I was thinking of `readonly` and `featured`. Then all Boolean properties have the same behavior.

icebob:
Ok, good idea.

icebob:
@mpociot could you make a new PR with `readonly` and `featured`?

PR #107, opened 2017-01-24
Add support for readonly and featured field functions

mpociot:
As discussed in #103 this PR adds support to use a callback function to modify the behavior of the "featured" and "readonly" field.


icebob:
Thank you!

Issue #112, opened 2017-02-12
Huge component size


cristijora:
Just noticed that the overall component has 400+ kb in size. 
I must say that's pretty huge for a component that generates fields.
After taking a look at the source code I noticed that `moment.js` is used in several places for very basic functionalities. I think you can get rid of it to save some space.

Only by removing all moment js dependencies (4) the size shrinked from 400kb to 176kb. 
There is also one component `fieldVueMultiSelect` which for some reason uses the whole `Vue` library. I don't think there is need to bring the whole library just to instantiate that component. There should be another way of doing it. By removing the vue dep the final file further reduces to 126kb.

Taking down `lodash` util function brings the file down under 100kb.
To sum it up, by removing only external dependencies (`moment, vue and lodash`) I managed to bring the library down from 400+ kb to 55kb and by removing external components it goes down to 29kb. 

The size here is huge factor for people deciding whether to use such a component or not. I would say we could shrink this one down to 100-150kb very easily and then with some effort it can be brought under 100kb.

icebob:
Thank you for your explain. Could you make a PR with this changes?

cristijora:
Yes, I will make one today with some simple changes which should reduce the size in half :)

icebob:
This is the current bundle stat:
![image](https://cloud.githubusercontent.com/assets/306521/22875954/96cdf20a-f1ce-11e6-93f6-1c27ab7d9b5e.png)
The source is only 2.8%. The others: vue (28%), moment (35%), lodash (42%)

cristijora:
Interesting. What did you use to get those stats? 
![image](https://cloud.githubusercontent.com/assets/15955045/22876059/828fd55c-f1d7-11e6-9f23-e8d7efb5a94e.png)

These are the 2 files you get when installing the component via npm (402 kb)

icebob:
I used this tool: https://chrisbateman.github.io/webpack-visualizer/
If you make a build, there will be a stats.json in root of repo. And you need to drop this file to the visualizer site.

lionel-bijaoui:
Hi, this issue has the same goal as my other one #67 
I absolutely agree with @cristijora and at the time, I already got impressive result.

I think the generator should be on it's own and people should just pick and choose which one they want.
I don't know if each field should have it's own sub repo (eg. vfg-checkbox) or if we could provide the same flexibility with only one repo.

What are your thought ?

cristijora:
[...]

PR #113, opened 2017-02-13
#112 Get rid of Vue dependency in the code.

cristijora:
Improve output bundle size by removing Vue dep and using moment.min.js
Reduced bundle size from 402kb to 176kb.

I think it's better to check if `vue-multi-select` is registered globally rather than checking if it's present in the window. Most of the time you register vue components with `Vue.component...` globally and don't necessarily put them in the window.


icebob:
Nice PR. I did't know the `this.$root` stuff :)

lionel-bijaoui:
I'm going to repeat myself from #114 
> I don't agree with the changes for moment.js.
By using the already minified version of moment, we reduce the chance of the minifier of vfg to do it's best job. It has to minify normal code with minified code, and **I don't claim to know how it work**, but I think that's not the best.
Also, we loose the sourcemap of moment.js.

I guess it I'm really nitpicking, but since the code will be minified at some point, why not wait for that moment (no pun intended). Or maybe minifying already minified code make it even more minified, and if it is the case, I'm sorry for my ignorance.

And I might add, the use of `this.$root` is super neat. Really well done ! 👍

PR #114, opened 2017-02-13
Update dev example after #113

icebob:


lionel-bijaoui:
I don't agree with the changes for moment.js.
By using the already minified version of moment, we reduce the chance of the minifier of vfg to do it's best job. It has to minify normal code with minified code, and **I don't claim to know how it work**, but I think that's not the best.
Also, we loose the sourcemap of moment.js.

PR #126, opened 2017-02-15
change name to "vfg" instead of "vue-form-generator" for the exported files

lionel-bijaoui:
change name to "vfg" instead of "vue-form-generator" for the exported files.

Temporary fix for the ENV variable being undefined


package.json
```diff
@@ -3,11 +3,11 @@
   "version": "2.0.0-beta4",
   "description": "A schema-based form generator component for Vue.js",
   "homepage": "https://github.com/icebob/vue-form-generator",
-  "main": "dist/vue-form-generator.js",
+  "main": "dist/vfg.js",
   "scripts": {
     "prebuild": "npm run test",
-    "build:core": "webpack --progress --config webpack.build.config.js --define process.env.FULL_BUNDLE=false --output-filename=vue-form-generator-core.js",
-    "build:full": "webpack --progress --config webpack.build.config.js --define process.env.FULL_BUNDLE=true",
+    "build:core": "webpack --progress --config webpack.build.config.js --define process.env.FULL_BUNDLE=false --output-filename=vfg-core.js",
```
icebob: Please change both build script to `cross-env`

webpack.build.config.js
```diff
@@ -15,17 +15,20 @@ var loaders = [
 		"loader": "vue"
 	}
 ];
-var cssFileName = "vue-form-generator.css";
-if (!process.env.FULL_BUNDLE) {
-	var cssFileName = "vue-form-generator-core.css";
+var cssFileName;
+console.warn(process.env.FULL_BUNDLE)
```
icebob: Please remove console message

package.json
lionel-bijaoui: Let me explain a bit why i used `cross-env`. That way We have an explanation for later if I forget why i did that.

Here `--define process.env.FULL_BUNDLE=false` is the same as using `webpack.DefinePlugin` in the config.
By doing so, it make `process.env.FULL_BUNDLE` available in the code. What I didn't notice at first, is the it not yet defined in the config file.
So by using `process.env.FULL_BUNDLE` there, I get `undefined`. But by using `cross-env`, it is available in the config (but not in the code).
So I have to use both `cross-env FULL_BUNDLE=true` and  `--define process.env.FULL_BUNDLE=true` to make it work.
This is a ugly temporary solution that will be fixed when the lib goes to webpack 2 one day.

Anyway, I added `cross-env` on both commands like you asked me in the next commit.

webpack.build.config.js
lionel-bijaoui: Sorry, removed

lionel-bijaoui:
Thanks

Issue #128, opened 2017-02-21
Disable submit button


pimhooghiemstra:
I am a bit stuck with the submit button of a form. I would like to disable the submit button when validation fails and enable it again once all fields pass validation.

However, according to the docs we can use the disabled property in the schema for any field and using it on a submit button adds a 'disabled' attribute to the HTML, but on the outer div.form-group. I expected it on the input[type=submit]...

My idea was to use the onValidated callback for an individual field to toggle this enabled/disabled state on the submit button.

Is something like this possible?

Here is a fiddle showing the issue: https://jsfiddle.net/pimhooghiemstra/0mg1v81e/207/

icebob:
The `disabled` property can be function too.
Please try it:
```
{
              type:"submit",
              buttonText:"Click me to submit",
              onSubmit:this.submitForm,
              disabled() {
		return this.errors.length == 0;
   	     }
}
```

> or `this.$parent.errors.length` ?!?!

icebob:
Hmm, not working because the `this` not binded to the instance of field.

icebob:
I will fix it in the next release.

pimhooghiemstra:
Great! I have a workaround for the moment but it quickly becomes ugly i am afraid.

icebob:
@pimhooghiemstra the solution & fix is coming with #129

icebob:
If you updated, could you try this solution?
```js
{
              type:"submit",
              buttonText:"Click me to submit",
              onSubmit:this.submitForm,
              disabled() {
		return this.$parent.errors.length != 0;
   	     }
}
```

pimhooghiemstra:
I'll try this tomorrow

pimhooghiemstra:
@icebob I copied the code above for the submit button, but unfortunately this gives me an error.
![screen shot 2017-02-23 at 10 40 07](https://cloud.githubusercontent.com/assets/7228111/23253376/8e544456-f9b4-11e6-8f35-d244301a42b1.png)

If I check the Vue devtools, it doesn't seem there is a parent errors object?

![screen shot 2017-02-23 at 10 43 30](https://cloud.githubusercontent.com/assets/7228111/23253476/f6b7574a-f9b4-11e6-921b-6e68647151ee.png)


Then I tried the disabled() function to just return true, thinking the button would be disabled on page load but that was not the case.

Maybe I am missing something?

icebob:
Yes, there are some other bugs :) I doing a hotfix.

[...]

Issue #109, opened 2017-02-11
More control over the validation process


cristijora:
Great component, but I would suggesting having the validation process more open to the ones who use the form generator. 
After reading the docs and looking in the source cod I didn't find a good way of validating my generated form exactly when I want. The only implicit way was to declare a `submit` input which in many real cases doesn't have to be right bellow the inputs. 
Another approach would be to declare a `ref="myForm"` and call `this.$refs.myForm.validate()` in my custom form submit function. This approach doesn't seem clean to me and might get ugly when you have multiple forms.
I suggest maybe adding an event which will pass the `errors` array or boolean up to the parent component.
Example: 
```javascript
<vue-form-generator 
     @errors-changed="handleErrors">
</vue-form-generator>

//parent component
data:function(){
 return{
  isValidForm:false,
 }
},
methods:{
  handleErrors:function(errors){
    this.isValidForm = errors.length===0;
   },
  myCustomSubmit:function(){
    if(this.isValidForm){
      //proceed with my form submission
   }
 }
}
```
Or alternatively 
```javascript
<vue-form-generator 
     @on-validate="handleValidation">
</vue-form-generator>

//parent component
data:function(){
 return{
  isValidForm:false,
 }
},
methods:{
  handleValidation(isValid){
   this.isValidForm = isValid;
  }
}
```

Necessary code change
``` javascript
validate() {
  // console.log("Validate!", this.model);
  this.clearValidationErrors();

  each(this.$children, (child) => {
    if (isFunction(child.validate))
    {
      // console.log("Validate ", child.model)
      let err = child.validate();
      each(err, (err) => {
        this.errors.push({
          field: child.schema,
          error: err
        });
      });
    }
  });
  let isValid = this.errors.length == 0;
  this.$emit('on-validate',isValid );
  return isValid;
},
```

icebob:
Good idea. I support the second solution with "on-validate" event.

icebob:
Other improvements:

If `validateAfterChanged` true, the fields run `validate` and the errors put to the `this.schema.errors`.

- [ ] doesn't modify the schema. Put to the local `data` and call an `on-validate` event which listen the parent.
- [ ] formGenerator listen the children `on-validate` event and append the errors to the self `this.errors`. Plus call self `on-validate`.
- [ ] call the `disabled`, `required`, `visible` & `readonly` schema functions with field instance as `this` (#128), plus pass schema as second argument & field instance as third argument

icebob:
[...]

PR #129, opened 2017-02-21
Improve validations & bindings

icebob:
Concerned issues: #128 #109 

1. VFG component emits a `validated` event, if validation executed. Event parameters: `isValid: boolean, errors: Array`

Example:
```js
<vue-form-generator @validated="onValidated" />

...

methods:{
  onValidated(isValid, errors) {
   console.log("Validation result: ", isValid, ", Errors:", errors);
  }
}

```

2. Fields validation `schema.errors` issue is fixed. Now `errors` in the local data. The communication with parent is event-based (similar `@validated` solution)

3. Fix `this` in schema functions. Now `this` is always the instance of field. Parameters: `model, fieldSchema, field`

### DIFF
--- a/dev/full/app.vue
+++ b/dev/full/app.vue
@@ -15,7 +15,7 @@
       <div class="errors text-center">
         <div v-for="(item, index) in validationErrors" track-by="index" class="alert alert-danger">{{ item.field.label}}: <strong>{{ item.error }}</strong></div>
       </div>
-      <vue-form-generator :schema="schema" :model="model" :options="formOptions" :multiple="selected.length > 1" ref="form" :is-new-model="isNewModel" @model-updated="modelUpdated"></vue-form-generator>
+      <vue-form-generator :schema="schema" :model="model" :options="formOptions" :multiple="selected.length > 1" ref="form" :is-new-model="isNewModel" @model-updated="modelUpdated" @validated="onValidated"></vue-form-generator>
     </div>
     <div class="col-md-6">
       <pre v-if="model" v-html="prettyModel"></pre>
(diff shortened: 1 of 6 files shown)


pimhooghiemstra:
@icebob nice work!

A couple of quick questions if I may?

- For my project I installed the npm version of vue-form-generator. Now I'd like to use this new version. I understand it is not on npm (you guys wait for version 2.0.0 I presume?). Currently I just forked and cloned the repo (with latest changes) to my computer, then compiled it and copied the vfg.js, vfg-core.js, vfg-core.css and vfg.css to my project.
However, I am not sure how to proceed, do I just place those in my node_modules? Haven't tried it , but instead copied them in my own components directory and imported from there. Although this works I am wondering if this is the normal way to do this?

I also noticed that the 'fields' array has changed substantially from the version on npm. I found out because I opened the examples. Is there another way to get to know this (besides looking at the code directly?)

Thanks in advance. I realise that my questions are actually more generic and not specific to this repo...

icebob:
@pimhooghiemstra I made a new build, so dist files contains modifications from this PR. So you can try this   version on your project if you change the version of vfg in package json to 
```
"vue-form-generator": "icebob/vue-form-generator"
```
So `npm install` will download the latest committed files. 

Yes, fields list changed, separated to two group (core, optional), removed deprecated fields and dist files renamed. They are a break changes, but we are in beta state, so we don't increment the mayor version. 
If you update to a beta version, check the changelog like when you update to a new major version of a lib. The [changelog](https://github.com/icebob/vue-form-generator/blob/master/CHANGELOG.md) of the next beta release will contain the break changes with example migration codes.

_If we release the first stable v2 version, we will follow semver versioning again._

pimhooghiemstra:
Great! 

Just installed vfg from git instead of npm, did know that it was possible, not yet how to do it, thanks!

I'll check changelog when the new beta version comes out, definitely!

icebob:
Only use this install method for testing, prototyping because it is always download the latest committed files, which can be unstable too.

pimhooghiemstra:
I understand, appreciate the warning! I'll wait for the 2.0.0 release for production anyways ;-)

PR #150, opened 2017-03-10
Fix a problem with value define but not an Array

lionel-bijaoui:
If the model is initialized with an empty string, when checked/clicked, the push was failing because value is not an array.
```js
model: {
    checklist:''
},
schema: {
    fields: [
            {
            type: "checklist",
            label: "CHECKLIST",
            model: "checklist",
            listBox: true,
            values: [{
                name: "HTML5",
                value: "HTML5-123"
            }, {
                name: "Javascript",
                value: "Javascript-123"
            }],
            }
	]
}
```

````
TypeError: this.value.push is not a function
````

### DIFF
--- a/src/fields/core/fieldChecklist.vue
+++ b/src/fields/core/fieldChecklist.vue
@@ -68,12 +68,13 @@
 			},
 
 			onChanged(event, item) {
-				if (isNil(this.value))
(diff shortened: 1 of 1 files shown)


lionel-bijaoui:
I found out that it doesn't really fix the problem. Another PR is coming.

Issue #145, opened 2017-03-07
Discrepancy, minor mistake, and inconsistency in the documentation and between similar functionnality


lionel-bijaoui:
`radios`'s `radiosOptions` is mandatory but flexible, and `checklist` is optional but strict.

While `checklist` support `values` as an array of string or an array of objects, `radios` only support the object format if `radiosOptions` is defined.
If you forget to add `radiosOptions`:
```
TypeError: this.schema.radiosOptions is undefined
```
What should be going on here in my opinion:
- both `radios` and `checklist` should take simple array or array of objects as `values` (`name` for display and `value` for internal/model value)
- both `radios` and `checklist` should have options (`radiosOptions` and `checklistOptions`) to define the name of each key used by the object

Also when using the object the [documentation say](https://icebob.gitbooks.io/vueformgenerator/content/fields/checklist.html#checklist-field-with-object-values) that it should save only the id in the model, but it save everything:
```js
// what it does
"checklistlistbox":
[
    {
        "name": "Javascript",
        "value": "Javascript-123"
    }
]
// instead of what the doc say it should do
"checklistlistbox": ["Javascript-123"]
```
None of this is very bad by itself, but I think core fields should be more coherent in their options and functionalities. It will help users and it will help us making unit test without headaches.

So what should be the basic behavior of these kind of fields ? `***Options` mandatory to use array of object ? Basic array and array of objects (with `name` and `value`) available, and only use `***Options` to customize the name of keys ?

If you have found other small problems like those, please leave a comment !
We should fix them before the v2.0.0

icebob:
I think both of them should work with basic array & object arrays. And should create a `checklistOptions` for `checklist` that we can set the name of keys.

icebob:
is it ready?

lionel-bijaoui:
I didn't had the time to check all the documentation. Some example may need an update for the v2.
From the question we got in the issues, I think we need more details about the visible/disable function parameters (#149), how to use VFG without webpack (#148), how to use the default value (#147) and how to translate the error messages (#134).
Maybe the documentation need a FAQ section ?

icebob:
Good idea. For this questions would be better a FAQ. 
Could you open a new issue about it?

PR #151, opened 2017-03-10
Move some dependency to devDependency

### LINKED ISSUE #135: Should `vue` be a `devDependencies` instead of a `dependency` ?
When I install VFG for my project, I realized that a different version of `vue` was installed. I guess npm does that because it it listed under `dependencies` in `package.json`.
But this is useless, no ?
If I want to clone the project to work on it, I will install `dependency` and `devDependencies`.
However, if I want to use VFG as a plugin/component, I don't need to install this specific version of `vue`.
I can do:
```
import {component} from 'vue-form-generator';

export default {
    components: {
        'vue-form-generator': component
    }
}
```
As long as I use vue 2 it should work

lionel-bijaoui:
close #135

### DIFF
--- a/package.json
+++ b/package.json
@@ -49,6 +49,7 @@
     "babel-plugin-transform-runtime": "6.23.0",
     "babel-preset-es2015": "6.22.0",
     "babel-preset-stage-2": "6.22.0",
+    "babel-runtime": "6.23.0",
     "chai": "3.5.0",
     "conventional-changelog-cli": "1.2.0",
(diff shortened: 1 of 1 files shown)


lionel-bijaoui:
You are welcome 😄

icebob:
You are very productive! ;)

lionel-bijaoui:
I'm doing my best 😅 
Using the plugins in a real project allow to identify issues faster and I need to fix them fast too :)
I'm going to work on #145 now

PR #152, opened 2017-03-10
Better behavior for radios and checklist

lionel-bijaoui:
#145
Now both `radios` and `checklist` support array of string array of string or array of objects (with `name` and `value` properties) by default.
If respectively `radiosOptions` or `checklistOptions` is defined, these key can be replaced by other name, but this is no longer mandatory to use the array of object notation.
The test are updated to test that behavior.


lionel-bijaoui:
I'm working on the documentation

Issue #110, opened 2017-02-12
Add support for custom validation messages


cristijora:
The component currently has a set of predefined validation error messages inside `resources` object in `validators.js`.
It would be very nice to be able to specify custom validation messages other than those that are predefined.
Example:
``` javascript
 required:true,
 min:6,
 validator: VueFormGenerator.validators.string,
 validatorMessage: "Password must be at least {1} characters",
```
This would basically replace the predefined message `textTooSmall` and would display the specified validatorMessage.

icebob:
It is not easy, because you can set multiple validators as an array. And some validators has multiple messages too.

cristijora:
I see your point.
Maybe have the possibility to specify the validators as an array of objects
``` javascript
validator:[{validator:VueFormGenerator.validators.string,message:"Password must be at least {1} characters"}]
```
Since you iterate through each validator in the array, you can verify if it's an object and if it has a custom message, then you can pass the custom message to the validator function. 
Don't know if it's worth it though after looking through the code.

icebob:
Thanks your suggestion, but I think, the syntax is too complex. I think, it would be more readable:

Current:
```js
{
	type: "text",
	min: 6,
	validator: validators.string
}
```

The new with custom locales: 
```js
{
	type: "text",
	min: 6,
	validator: validators.string.locale({ textTooSmall: "Password must be at least {1} characters" })
}
```
You can add more messages too. The "name" of message is come from the currently used [validators.resources](https://github.com/icebob/vue-form-generator/blob/master/src/utils/validators.js#L22) object.
And for this only need a wrapper function for validators to create the "locale" func.

What do you think?

cristijora:
Well it is indeed more readable and doesn't require too many changes but it implies that the developer has to look at the source code and identify when certain `locale` messages are return and which one to replace. 
In the end it's better  than it is now, so I agree with the new possible implementation

icebob:
I implemented it in #169

PR #169, opened 2017-03-23
Custom validation for fields

icebob:
It is the implementation of #110 .

You can add custom validator messages for fields in schema.
Every validator function has a method `locale`. Parameter is an object of custom messages. 

Example:
```js
let schema = {
	fields: [
		{
			type: "input",
			inputType: "password",
			label: "Password",
			model: "password",
			min: 6,
			required: true,

			// String validator with custom error messages
			validator: validators.string.locale({ 
				fieldIsRequired: "The password is required!",
				textTooSmall: "Password must be at least {1} characters"
			})
		}
	]
}
```

Result:
image


icebob:

Issue #170, opened 2017-03-28
Async validators


jmverges:
Hi folks,

I'm trying to do this:

```
Vue.use(VueFormGenerator);
    VueFormGenerator.validators.short = function (value, field) {
        if (value.length < 3) {
            return ['The short must have at least 3 characters'];
        }

        return axios.post('api/scaffolder/validate-short', {"short": value}).then(response => {
                    if (response.data.valid == false) {
                        return ['The short already exists'];
                    }
                    return null;

                });
    };
```

However if the value is > 2 field is always in green, even if response.data.valid is false. Is VFG able to have ajax validations?

Best,
Juan Manuel.

cristijora:
Can you post the usage of your validator ?

This part seems a bit wrong to me since your are returning a Promise here basically instead of an array
```js
return axios.post('api/scaffolder/validate-short', {"short": value}).then(response => {
                    if (response.data.valid == false) {
                        return ['The short already exists'];
                    }
                    return null;

                });
```

icebob:
vfg can't support async validators.

jmverges:
@icebob really?

jmverges:
@icebob but would you like to support them? if not I couldn't use this repo anymore

jmverges:
what about drop de built in validation, change the fields generation and integrate with Vee Validate?

cristijora:
Dont think that's an option. For example I got rid of Vee validate and all it's stuff and switched to form generator and form generator validations because Vee validate requires a lot of boilerplate code. If you want a field validation you have to add like 3 html wrappers and functions and it gets really messy with custom components like date pickers or external stuff you cannot control so easily.

jmverges:
yep, that's why I bet for VFG when I saw it, however, ajax validation is a must for me as is the most common use case in my world. Not having this feature implemented here makes the library less powerful (no offence). I could spend a bunch of hours working on this feature if @icebob wants to support async validation and he has an idea about which path to take for this one.

cristijora:
Does vee validate support async validation?

jmverges:
@cristijora yes, it works quite well.

cristijora:
Can you post a fiddle so I understand how it works?

icebob:
@jmverges I think it is a good feature.

icebob:
I'm trying a quick implementation with Promise

cristijora:
Please share the result/attempt in a PR maybe.

icebob:
[...]

PR #171, opened 2017-03-28
support async (promised) validators

icebob:
Handle async validator which returns with Promise. Discussion: #170 

Example async validators in dev:

```js
let customAsyncValidator = function(value) {
	return new Promise((resolve, reject) => {
		setTimeout(() => {
			if (value)
				resolve();
			else
				resolve([ "Invalid value from async validator" ]);
		}, 1000);
	});
};
```

Schema:
```js
{
	type: "input",
	inputType: "url",
	label: "URL",
	model: "website",
	validator: customAsyncValidator
}
```

TODO:
- [ ] add tests


cristijora:
Nice. @jmverges take a look :)

jmverges:
@icebob works properly for me, thank you!

icebob:
Great! You're welcome.

icebob:

Issue #172, opened 2017-03-29
File upload


fizerkhan:
I want to upload CSV file. What is the input type for file upload?

cristijora:
I don't see file upload being handled in the code. You can go with your own [custom component](https://icebob.gitbooks.io/vueformgenerator/content/fields/custom_fields.html) and use any of the exisiting file upload components that are out there (e.g [here](https://github.com/lian-yue/vue-upload-component) )

fizerkhan:
@cristijora Your commit will fix the issue right?

cristijora:
Yes, the [PR](https://github.com/icebob/vue-form-generator/pull/173) got merged and it will be available in the next build or if you need it now simply use the [js from here](https://github.com/icebob/vue-form-generator/tree/master/dist)

fizerkhan:
@cristijora Thanks. 👍

fizerkhan:
@cristijora I got this error after integrating this js file

```
DOMException: Failed to set the 'value' property on 'HTMLInputElement': This input element accepts a filename, which may only be programmatically set to the empty string.
    at Array.updateDOMProps (http://localhost:3000/app/app.js:10043:19)
    at patchVnode (http://localhost:3000/app/app.js:9559:62)
    at updateChildren (http://localhost:3000/app/app.js:9480:9)
    at patchVnode (http://localhost:3000/app/app.js:9564:29)
    at VueComponent.patch [as __patch__] (http://localhost:3000/app/app.js:9689:9)
    at VueComponent.Vue._update (http://localhost:3000/app/app.js:7663:19)
    at VueComponent.<anonymous> (http://localhost:3000/app/app.js:7630:10)
    at Watcher.get (http://localhost:3000/app/app.js:6827:27)
    at Watcher.run (http://localhost:3000/app/app.js:6896:22)
    at flushSchedulerQueue (http://localhost:3000/app/app.js:6710:13)
logError @ app.js:5601
```

cristijora:
I will look into it.

cristijora:
@fizerkhan Try not to set the initial model value for the file input to an empty string. Does this appear every time you upload something or only initially?

fizerkhan:
@cristijora It happens only after I select a file from the file dialog.

fizerkhan:
@cristijora I fixed this by setting file value to ''.

```
this.model.file = '';
```

BramVer:
[...]

PR #173, opened 2017-03-29
File upload #172

cristijora:
After looking in the code and testing stuff out I didn't find any way to get file data from file input type.
Added a `change` listener and set the model `value` to the files list in case an input is of file type.


icebob:

Issue #197, opened 2017-05-16
Is fieldset required for VFG ?


lionel-bijaoui:
Since VFG is using a `<fieldset>` as main tag, it is [impossible to use flexbox on it](http://stackoverflow.com/questions/28078681/why-cant-fieldset-be-flex-containers).
Flexbox allow for great layout to be made without any CSS hacks.
Knowing that most behaviors of forms are dealt with through use of JavaScript, can we simply use a div as a "wrapper" for the fields ?
Would that break something essential that I'm missing to see ?

lionel-bijaoui:
I have a first solution to make it a little more flexible with the `is` attribute.
```html
// vfg.vue
<template>
    <div :is="tag">VFG</div>
</template>

<script>
export default {
    props: {
        tag: {
            type: String,
            default: 'fieldset'
        }
    }
};
</script>
```

```html
basic usage
<vfg></vfg>
config usage
<vfg tag="section"></vfg>
```

```html
basic usage rendered
<fieldset>VFG</fieldset>
config usage rendered
<section>VFG</section>
```

DelfsEngineering:
Can we revisit this? flexbox is the jam and we simply can't use it because we cant attach a class to the field wrapper. We really need to be able to add classes or better, remove the `fieldset` wrapper completly.

zoul0813:
@DelfsEngineering check formGenerator.vue

There’s a “tag” property you can assign to field groups, just use the field groups feature and you can also use the field group classes logic as well.

PR #198, opened 2017-05-16
Add a props to change the main tag

lionel-bijaoui:
#197

### DIFF
--- a/src/formGenerator.vue
+++ b/src/formGenerator.vue
@@ -73,9 +73,12 @@ div
 			},
 
 			tag: {
-	            type: String,
-	            default: 'fieldset'
-	        }
+				type: String,
+				default: "fieldset",
+				validator: function (value) {
+					return value.length > 0;
+				}
+			}
 		},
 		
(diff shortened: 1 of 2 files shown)


icebob:
Great! But please add relevant test too.

lionel-bijaoui:
@icebob I don't really know what to test... `is` is a vuejs feature and I don't know where to begin here. Can you give me a hand ?

icebob:
We [here](https://github.com/icebob/vue-form-generator/blob/master/test/unit/specs/VueFormGenerator.spec.js#L46) test the root element is a "fieldset". Clone it, set the `tag` prop and check the result :) Thanks!

lionel-bijaoui:
Oh ok ! I feared I needed to test everything, what a relief 😅 
I do it right now ! Thanks !

lionel-bijaoui:
@icebob I have beefed up the props with a simple validator. If you feed `tag` with an empty value, it will fail visibly.
Good to merge ?

PR #205, opened 2017-05-25
Improve schema defaults

dflock:
Supply default values for `schema.type`, `schema.inputType` & `schema.model`.

This is a test, to see what you think, but the idea is to make schema's smaller by supplying defaults for some fields if they're not supplied. This would make this:

```js
schema: { 
  fields: [
    {
      type: "input",
      inputType: "text",
      label: "First Name",
      model: "first_name",
    },
    {
      type: "input",
      inputType: "text",
      label: "Last Name",
      model: "last_name",
    }
  ]
}
```

equivalent to this:

```js
schema: { 
  fields: [
    { label: "First Name" },
    { label: "Last Name" }
  ]
}
```

... which for large forms with lots of text boxes, ends up being a huge saving.

This works by adding a `beforeMount()` to `formGenerator.vue` which sets defaults in the `schema`, if required. I also added a `dev/defaults/index.html` test for this.

This works, as far as it goes - but it seems to break about 50% of the tests, for reasons I haven't looked into.

I'm not sure about this particular implementation, but I _do_ think that making the schema declarations smaller for the common case would be a good thing.

### DIFF
--- a/dev/defaults/app.vue
+++ b/dev/defaults/app.vue
@@ -1,0 +1,36 @@
+<template lang="html">
+	<div>
+		<vue-form-generator :schema="schema" :model="model" :options="formOptions"></vue-form-generator>
+	</div>
+</template>
+
+<script>
+import Vue from "vue";
+import Multiselect from "vue-multiselect"
+Vue.component("multiselect", Multiselect);
+
+export default {
+	data () {
+		return {
+			model: {
+				skills: ["Javascript", "VueJS"],
+				first_name: "Brian Blessed"
+			},
+
+			schema: {
+				fields: [
+					{
+						label: "First Name"
+					}
+				]
+			},
+
+			formOptions: {}
+		}
+	},
+
+	created() {
+		window.app = this;
+	}
+}
+</script>
--- a/dev/defaults/index.html
+++ b/dev/defaults/index.html
@@ -1,0 +1,12 @@
+<!doctype html>
+<html lang="en">
+  <head>
+    <meta charset="utf-8">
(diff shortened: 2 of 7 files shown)


dflock:
I also pulled the `slugify` stuff out of `abstractField.getFieldID()` and moved it to it's own file, so I could re-user it. I should probably put that stuff into a separate pull request.

icebob:
The idea is good. My problem is that model name generation is very tied. For example i don't use property names as `model.first_name`. I'm using camel-case names as `˛model.firstName`. 

@lionel-bijaoui What do you think?

jmverges:
Hi guys,

what about to have a property where you can set the callback for a custom transformation or one of the following string names and provide internally snake_case, kebap-case, lowerCamelCase and UpperCamelCase?

icebob:
It's better.

lionel-bijaoui:
Hi @icebob ! Honestly, for this kind of things, I think it's better to create your own "parser" that prepare the schema for you. This is an optimization that may be integrated in the future, but right now, for the v2, I think we should lock the functionality of VFG, fix all the bugs and get out of beta.
I'm not saying this is a bad idea, just that it is breaking a lot of internal system, for a return on investment that is really small and at a bad timing.

crashkonijn:
How about instead of providing the label, you provide the model and generate the label?

A title_case() of firstName, FirstName, first_name, first-name would all result in the label "First Name"

dflock:
That actually seems like a really good, rather simple way to satisfy people's objections to the original pull request - I wish I'd thought of it that way originally!

PR #206, opened 2017-05-25
Add support for schema.legend & field.id prefixes

dflock:
1. Add support for an optional legend for each schema/fieldset. Expects the schema to look something  like this:

```
schema: {
  legend: "Contact Details",
  fields: [
    {
      type: "input",
      inputType: "text",
      label: "Name",
      model: "name"
    },
...
```

2. Add support for field id prefixes, at the form level. So you can add a `fieldIdPrefix: 'prefix_here_'` to your `formOptions` and this will be prepended to _all_ field id's. So if you have this:

```js
schema: {
  legend: "Contact Details",
  fields: [
    {
      type: "input",
      inputType: "text",
      label: "Name",
      model: "name"
    },
    {
      type: "input",
      inputType: "email",
      label: "Email",
      model: "email",
      id: "user_email"
    }
  ]
},


formOptions: {
  fieldIdPrefix: 'frm1_'
}
```

you'll get this:

```html
<fieldset class="vue-form-generator">
  <legend>Contact Details</legend>
  <div class="form-group field-input">
    <label for="frm1_name">Name</label>
    <div class="field-wrap">
      <div class="wrapper">
        <input id="frm1_name" type="text" class="form-control">
      </div>
    </div>
  </div>
  <div class="form-group field-input">
    <label for="frm1_email">Email</label>
    <div class="field-wrap">
      <div class="wrapper">
        <input id="frm1_email" type="email" class="form-control">
      </div>
    </div>
  </div>
</fieldset>
```

So the `idPrefix` is (intentionally) applied to both auto-generated and manually set field ids.

I added a `dev/multipleforms` thing to demo this - which should probably be called `multiplefieldsets` or `legends` or something.

### DIFF
--- a/src/formGenerator.vue
+++ b/src/formGenerator.vue
@@ -1,6 +1,7 @@
 <template lang="pug">
 div
 	fieldset.vue-form-generator(v-if='schema != null', :is='tag')
+		legend(v-if='schema.legend') {{ schema.legend }}
 		template(v-for='field in fields')
 			.form-group(v-if='fieldVisible(field)', :class='getFieldRowClasses(field)')
 				label(v-if="fieldTypeHasLabel(field)", :for="getFieldID(field)")
@@ -119,6 +120,13 @@ div
 							this.clearValidationErrors();
 					});
 				}
+			}
+		},
+
+		beforeMount() {
+			// Add idPrefix to fields if fieldIdPrefix is set
+			for (let field of this.schema.fields) {
+				field.idPrefix = this.options.fieldIdPrefix || "";
 			}
 		},
 
--- a/dev/multipleforms/app.vue
+++ b/dev/multipleforms/app.vue
@@ -1,0 +1,59 @@
+<template lang="html">
+	<form>
+		<vue-form-generator :schema="section1" :model="model" :options="formOptions"></vue-form-generator>
+		<vue-form-generator :schema="section2" :model="model" :options="formOptions"></vue-form-generator>
+	</form>
+</template>
+
+<script>
+import Vue from "vue";
+
+export default {
+	data () {
+		return {
+			model: {
+				name: 'Brian Blessed',
+				email: "brian@hawkman.mongo",
+				pref_1: 'blah'
+			},
+
+			section1: {
+				legend: "Contact Details",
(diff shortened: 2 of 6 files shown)


icebob:
I will check it.

jmverges:
👍

jmverges:
One thing, regarding this,  

should we then use several <vue-form-generator></vue-form-generator> one for each legend? 

I would check if in schema exists a property called **groups** which is an array with objects like {legend:'Something',fields:[]} and if that property doesn't exist then go fields.

Definitely, this kind of feature is needed but having more than one tag has no sense for me

Cheers,

src/formGenerator.vue
```diff
@@ -1,6 +1,7 @@
 <template lang="pug">
 div
 	fieldset.vue-form-generator(v-if='schema != null', :is='tag')
+		legend(v-if='schema.legend') {{ schema.legend }}
```
jmverges: I would add a bit more of logic in here to not need more than one tag <vue-form-generator></vue-form-generator>

src/formGenerator.vue
```diff
@@ -122,6 +123,13 @@ div
 			}
 		},
 
+		beforeMount() {
+			// Add idPrefix to fields if fieldIdPrefix is set
+			for (let field of this.schema.fields) {
```
jmverges: same here

icebob:
@jmverges Yes, it would be good, because the grouping is an [old todo](https://github.com/icebob/vue-form-generator/projects/1#card-210134) :+1:

jmverges:
@dflock are you gonna take care or would you prefer if I fork your branch and do the stuff?

@icebob I have a question, it is possible to update the schema and rebuild it once is created?

icebob:
If you change the schema the Vue will be rebuild the DOM. (I think)

jmverges:
I think is not working for me

jmverges:
@dflock @icebob I have a groups versions locally working. Should I create a new PR?

icebob:
Wow, you are very fast :) Yes, please open a new PR

icebob:
Closed by #209

Issue #253, opened 2017-07-12
Validator: Field validated class


lionel-bijaoui:
Hi @icebob @dflock @cristijora, 

I need a validated state (class) for my fields.
I noticed [in the source](https://github.com/icebob/vue-form-generator/blob/master/src/formGenerator.vue#L177) that there is a `validationSuccessClass` and by setting it with `validationErrorClass`, I can get a class that currently correspond to "no error".

But "no error" and "validated" don't have the same meaning.
A field should have 3 states : "normal", "error" and "validated".

With `validateAfterLoad: false, validateAfterChanged: false`, I want my field to be "normal" until I trigger a validation. Then all fields are either "error" or "validated".

With `validateAfterLoad: false, validateAfterChanged: true`, I want my field to be "normal" until I edit the field. Then this field is either "error" or "validated".

With `validateAfterLoad: true, validateAfterChanged: false`, I want all my fields to be either "error" or "validated" after initial load. They keep their state until I trigger a validation.

With `validateAfterLoad: true, validateAfterChanged: true`, I want all my fields to be either "error" or "validated" after initial load. They keep their state until I edit the field.

With this behavior, it will be possible to add a checkmark or anything I want next to each "validated" field. It will allow my users to see their progress on the from.

TL:DR The goal is to have a class that mean "I have been through the validation process at least once and I have no error"

How can we achieve that ?
I can do a PR if necessary but I wanted your opinion on the issue first.

Thank you !

icebob:
Good idea. What is your suggestion?

cristijora:
@lionel-bijaoui Totally agree that there should be a another state which is the "initial" state before validating a field.
I don't think the api should change (validationSuccessClass, validationErrorClass are fine as they are)
How do you want to set this intermediate state ? I looked real quick and I don't think there is a way to distinguish between a field with no errors and a field that has been validated and has no errors.

lionel-bijaoui:
I was thinking of using `abstractField.js` to add a data property `vfg_ValidationState` that track if the field have been run through the validator. At first, it is `false` and if the method `validate` is run, this property change to `true`.
The next change is in `formGenerator`, the method `getFieldRowClasses` need to change to use that new property to decide if it can apply an `error` or `success` class or no class at all.
```js
// formGenerator.vue
methods: {
    // Get style classes of field
    getFieldRowClasses(field) {
        const hasErrors = this.fieldErrors(field).length > 0;
        let baseClasses = {
[...]

PR #257, opened 2017-07-13
Introduce a new state for validation class.

lionel-bijaoui:
#253 
This is copying the way errors works.
While it work, I'm sure their is a better way to do validation.
I don't know yet how to change the test to pass, but since this may change I'm not going to change it for now.
Thank you for any feedback !

### DIFF
--- a/src/formGenerator.vue
+++ b/src/formGenerator.vue
@@ -358,8 +358,7 @@ div.vue-form-generator(v-if='schema != null')
 			},
 
 			fieldTouched(field) {
-				let res = this.touched.filter(e => e.field == field);
-				return res.map(item => item.touched)[0];
+				return this.touched.find(e=>e.field == field);
 			},
 
 			getFieldID(schema) {
--- a/src/formGenerator.vue
+++ b/src/formGenerator.vue
@@ -76,7 +76,7 @@ div.vue-form-generator(v-if='schema != null')
 						validateAfterLoad: false,
 						validateAfterChanged: false,
 						validationErrorClass: "error",
-						validationSuccessClass: "",
+						validationSuccessClass: "success"
 					};
 				}
 			},
@@ -102,6 +102,7 @@ div.vue-form-generator(v-if='schema != null')
 
 		data () {
 			return {
+				touched: [],
 				errors: [] // Validation errors
 			};
 		},
@@ -165,8 +166,10 @@ div.vue-form-generator(v-if='schema != null')
 			// Get style classes of field
 			getFieldRowClasses(field) {
 				const hasErrors = this.fieldErrors(field).length > 0;
+				const wasTouched = this.fieldTouched(field);
 				let baseClasses = {
-					error: hasErrors,
+					error: hasErrors && wasTouched,
+					success: !hasErrors && wasTouched,
 					disabled: this.fieldDisabled(field),
 					readonly: this.fieldReadonly(field),
 					featured: this.fieldFeatured(field),
@@ -174,13 +177,13 @@ div.vue-form-generator(v-if='schema != null')
 				};
 
 				let {validationErrorClass, validationSuccessClass} = this.options;
-				if (validationErrorClass && validationSuccessClass) {
+				if (validationErrorClass && validationSuccessClass && wasTouched) {
 					if (hasErrors) {
 						baseClasses[validationErrorClass] = true;
 						baseClasses.error = false;
-					}
-					else {
+					} else {
 						baseClasses[validationSuccessClass] = true;
+						baseClasses.success = false;
 					}
 				}
 
@@ -283,6 +286,12 @@ div.vue-form-generator(v-if='schema != null')
 			onFieldValidated(res, errors, field) {
 				// Remove old errors for this field
 				this.errors = this.errors.filter(e => e.field != field.schema);
+				this.touched = this.touched.filter(e => e.field != field.schema);
+
+				this.touched.push({
+					field: field.schema,
+					touched: true
+				});
 
 				if (!res && errors && errors.length > 0) {
 					// Add errors with this field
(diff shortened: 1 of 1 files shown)


src/formGenerator.vue
```diff
@@ -342,6 +357,11 @@ div.vue-form-generator(v-if='schema != null')
 				return res.map(item => item.error);
 			},
 
+			fieldTouched(field) {
+				let res = this.touched.filter(e => e.field == field);
```
cristijora: you could use
`let field = this.touched.find(e=>e.field == field)`
which returns null or the first match. This way you don't have to use `res.map(item => item.touched)[0];` which is not very safe

src/formGenerator.vue
```diff
@@ -283,6 +286,12 @@ div.vue-form-generator(v-if='schema != null')
 			onFieldValidated(res, errors, field) {
 				// Remove old errors for this field
 				this.errors = this.errors.filter(e => e.field != field.schema);
+				this.touched = this.touched.filter(e => e.field != field.schema);
+
+				this.touched.push({
```
cristijora: Why push touched fields every time a validation occurs? 
Maybe have a logic
```js
if( findTouchedField(field) == null)
  this.touched.push...

```

src/formGenerator.vue
```diff
@@ -301,10 +310,16 @@ div.vue-form-generator(v-if='schema != null')
 			// Validating the model properties
 			validate() {
 				this.clearValidationErrors();
+				this.touched.splice(0);
 
 				this.$children.forEach((child) => {
 					if (isFunction(child.validate))
 					{
+						this.touched.push({
```
cristijora: Just wondering if it's not easier to have `touched` inside field `data`, initialize it with false and when you call this validation method you simply push to this array that value. This should happen only upon first validation

lionel-bijaoui: In fact, I started with this strategy (to have touched inside field data), but I didn't think about doing what you suggest.
The problem is that the style is decided `getFieldRowClasses` (in `formGenerator.vue`). This method argument is the field definition from the schema. This field definition is what is used to match "local" errors with "global" errors (`fieldErrors` for example).
I don't know why @icebob used this way to match things, but I can assume that there is a reason for it, a consequence I don't foresee right now (maybe for groups ?).
Anyway, I'm going to try another way to match fields for `touched` property.

src/formGenerator.vue
lionel-bijaoui: Yes sorry, I was tired when I did that and did not bother to write good code. Thank you !

src/formGenerator.vue
cristijora: Well the idea would be that you still have this `touched` array only that the touched property is kept inside the field just like errors https://github.com/icebob/vue-form-generator/blob/master/src/fields/abstractField.js#L27
You would keep `touched` inside abstractField data and change upon validation of the field.

lionel-bijaoui: Yes, I understand, as I said, this was my first idea. But how do you access `touched` from `formGenerator` ?
I was going to add an index to the v-for loop (`(field, index) in fields`), but I noticed that groups will break this (I just noticed groups are not working how I assumed they would and this is another problem).
Anyway, changing an internal `touched` property is not the problem, accessing it from the parent (`formGenerator`) is the difficulty. How to match the children with the parent ?

cristijora: Just like it's done with errors. Iterate over children 
Another, more cleaner way IMHO would be to have something like this:
Child:
```js
mounted(){
      this.$parent.addField(this)
    },
    destroyed() {
        if (this.$el && this.$el.parentNode) {
            this.$el.parentNode.removeChild(this.$el);
        }
        this.$parent.removeField(this);
    },
```

Parent
```js
addTab(item) {
            const index = this.$slots.default.indexOf(item.$vnode);
            this.fieldData.splice(index, 0, item);
        },
        removeTab(item) {
            const index = this.fieldData.indexOf(item);
            if (index > -1) {
                this.fieldData.splice(index, 1);
            }
        },
```

But that is probably out of scope now.

lionel-bijaoui: What is `$vnode` ?

cristijora: It's a special data type/class which represents a component or a tree of elements in vue
https://github.com/vuejs/vue/blob/dev/src/core/vdom/vnode.js
It basically contains the element, tag, component data and many other things. Since slots return a list of `$vnodes` then you can do the extra check and see if the component you are trying to add exists in slots if slots are used. Since vfg uses `component :is`, then the filtering probably can be done based on `$children` maybe

lionel-bijaoui: I feel so dumb right now 😅 
I didn't now about this at all... My understanding of Vue is far from complete.

I like your solution a lot. Maybe I can introduce it only for this new state for now, and maybe move errors and other functionality later if no problem arise ?
If I understand correctly, I need to change that:
```js
addField(item) {
            const index = this.$slots.default.indexOf(item.$vnode);
            this.fieldData.splice(index, 0, item);
 }
```
into this:
```js
addField(item) {
            const index = this.$children.default.indexOf(item.$vnode);
            this.fieldData.splice(index, 0, item);
 }
```

cristijora:
```js
addField(item) {
            const index = this.$children.findIndex(field=> field.$vnode == item.$vnode);
            this.fieldData.splice(index, 0, item);
 }
```

lionel-bijaoui: Ok so your solution doesn't work.
When `validateAfterLoad` is true, the loop will trigger `getFieldRowClasses` before any field component initialization (`created` or `mounted`) and thus `fieldData` is empty, causing an error.
I think a solution like this would need a big rewrite of VFG internal.
I'm sorry but I cannot allow myself to lose any more time on this. I'm going to implement the fixes you suggested but I will keep this "dirty" for now.
Hope you understand, thank you for your help and feel free to do another PR with a better solution.

src/formGenerator.vue
lionel-bijaoui: Because it is cleaned at every validation

src/formGenerator.vue
cristijora: Sure no problem. Those were only suggestions, If they take too much time then any solution is ok

icebob:
@lionel-bijaoui Are you going to fix the errors? And make tests to cover the new property?

lionel-bijaoui:
Sorry I missed your approval. I'm working on that tomorrow.

robertbossaert:
Is this still being worked on? I can imagine this is a pretty nice feature to have; a lot of forms require success styling to be triggered when a user has really touched the field.

lionel-bijaoui:
I'm going to rework on this for the v3, closing this PR

PR #283, opened 2017-09-02
Proposal : Adding theming / layout functionalities for form generation

sleroy:
Hi, I think you have done right with your implementation of form generation.
However it is lacking a crucial feature to be really useful : allowing developers to build their own layouts / themes for the form generation like in frameworks such Laravel.

I have done a preliminary evolution of your code to show the e possibilities.

Basically I have created an abstractLayout and moved all the code of the for generator into it.
Secondly I have created a meta component form generation applying the render using a props 'theme'.
The form generation implemoentation has been renamed "form-bootstrap3" to express the fact that it's a form theme for Bootstrap 3.

What do you think about it ? Do you want to go forward ?


cristijora:
Hi @sleroy 

Although, I find this approach very interesting, the current PR adds quite a lot of formatting changes which makes it rather hard to observe changes + the changelog, readme changes which I think are not necessary.

For me, an ideal solution would be to specify the layout upon installation via scoped slots for example.
```js
Vue.use(VueFormGenerator, MyCustomLayout)
```

Where MyCustomLayout would be basically my custom definition of vfg layout.
```js
<form-generator-layout>
 <template scope="props">
  <div v-for="field in fields>
    //  ... my code
  </div>
</template>
</form-generator-layout>
```
This will allow other people to define layouts for different UI libraries which we can then import rather than define everything ourselves. 
However, this requires a very well defined set of "props" which should be minimal and shouldn't change over time.

icebob:
Thanks, the idea is good. But I won't merge it, because your PR contains a lot of not relevant changes.
Otherwise I think the slot-kind solution would be better, what @cristijora mentioned.

sleroy:
Hi @cristijora ,

Well the pull-request is rather a prototype than a real PR to be inserted.
Indeed the code is messy and I have been fighting with the actual code to clearly define what I was expecting of the framework.

Moreover I am rather new with vueJS and I did not know this scope templating functionality. I will have a look on that.

About the feature itself, 
- The library can provides some CSS fwk implementation but it should be a separate project.
-  CSS frameworks as bootstrap 3 is providing at least 2 or 3 ways to use forms. And applications usually are using both. (inline, groups ...)
-  Depending of the screen (modal, etc)  you may have to switch layouts.
- Fields are implemented as components (it looks a good idea) however it makes their HTML appearance complex to overload.

For example the Google adress field, I would like to increase the size (adding the input-lg class). How to do it cleanly ?

Anyway, I would like to thank you for the actual features. I am working on a fork for my project, but if the main project desires to implement the functionality, I am interested to give an hand.

PR #295, opened 2017-09-19
added css classes for core fields

vuchl:


icebob:
Any updates?

vuchl:
I updated the PR here #318. will try to fix travis errors now

Issue #276, opened 2017-08-24
[Vue-warn] when using custom fields


ndro:
Sampurasun!

Trying to create custom fields like [this](https://icebob.gitbooks.io/vueformgenerator/content/fields/custom_fields.html), it works perfectly. But, when I am looking on 'inspect element --> console' this [Vue-warn] was appear:

```
[Vue warn]: Property or method "formatValueToModel" is not defined on the instance but referenced during render. Make sure to declare reactive data properties in the data option.

found in

---> <FieldBulmaInput> at /var/www/vuejs/next-level-vuejs/client/helpers/custom-form-generator/fieldCustomInput.vue
       <VueFormGenerator>
         <TabContent>
           ...

```

It's happen when create 'input text' and 'textarea'. This is an example code of custom fields fieldCustomInput.vue:

```
<template>
<div class="field">
  <label class="label">{{ schema.customLabel }}</label>
  <div class="control">
    <input 
	class="input is-medium" 
	type="text"
	v-model="value"
	:disabled="disabled"
	:maxlength="schema.maxlength"
	:placeholder="schema.placeholder"
	:readonly="schema.readonly">
  </div>
  <br>
</div>

</template>

<script>
  import { abstractField } from 'vue-form-generator'

  export default {
    mixins: [ abstractField ]
  }
</script>

<style lang="scss" scoped>
.label {
	font-size: 1.5rem;
}
</style>

```

And I tried to defined 'formatValueToModel' in tag <script>, but it still didn't change anything.

Any idea what happen with my code?

Thanks

lionel-bijaoui:
It is not your fault, I think.
In [`abstract.js`](https://github.com/icebob/vue-form-generator/blob/master/src/fields/abstractField.js#L51), there is a check to see if `formatValueToModel` exist as a method. This is a way for custom field developer to format `value` how they want, if needed.
I think that check trigger `[vue-warn]`, but it is not a problem. In production, this message should not appear.
I'm going to look into it.

random-one:
I have the same issue. Trying to use datePicker from `vue-strap`. 
Whole field component code:
```
<template>
  <datepicker v-model="value" :disabled-days-of-Week="disabled" :placeholder="schema.placeholder">
  </datepicker>
</template>
<script>
import { abstractField } from 'vue-form-generator'
import { datepicker } from 'vue-strap'
export default {
  components: {
    datepicker,
  },
  mixins: [ abstractField ],
}
</script>
```

[...]

PR #298, opened 2017-09-22
Default formatValueToField and formatValueToModel implementations

alepeino:
Adds a default implementation for `formatValueToField` and `formatValueToModel` in the `abstractField` mixin. A simple identity function, which can just be overridden in the custom fields. 

A "tell, don't ask" approach, which makes the code simpler and more readable, in my opinion. Also, it's a fix for #276.


icebob:
Sorry, I didn't see it. I solved it in https://github.com/icebob/vue-form-generator/commit/33538605a5a82379d931fff59086cf5f981b7a8e

Thanks!

Issue #279, opened 2017-08-30
Switch styling doesn't work


GreymondTheLong:
I was trying to get the switch field working correctly. It renders as a normal checkbox and only had the correct styling after I included the css file [referenced on the jsfiddle demo](https://unpkg.com/vue-form-generator@2.0.0/dist/vfg.css)

icebob:
Could you make a jsfiddle?

GreymondTheLong:
I modified the original jsfiddle and removed the css file.

https://jsfiddle.net/gfg3Lmw4/

ekingery:
For anyone else who is having this issue, here is how I was able to resolve it in my `src/main.js`: 

`import "vue-form-generator/dist/vfg.css";`

lionel-bijaoui:
I'm sorry but I don't understand where is the problem.
If you remove the style, of course it will not be styled.
It fallback to checkbox but still work as intended.
Like @ekingery said, if you want the style, you need to import it.

ekingery:
I think the problem stems from the documentation, which never explicitly states that you need to import the CSS for the non-core package. I created a PR for this, hope that helps!

https://github.com/icebob/vue-form-generator/pull/305

PR #305, opened 2017-10-05
Clarified css import instructions

ekingery:
Related to / in response to #279


ekingery:
I didn't see the repo / docs for the gitbook, but those should probably be updated as well if this PR is accepted.

ekingery:
Thank you for the component, it's terrific!

PR #310, opened 2017-10-10
Use `name` in checklist input fields.

### LINKED ISSUE #243: Form inputs are not having 'name' attributes
While submitting all details, I wanna do it via ajax and json encoded code. Also i don't know how many fields i would have in each form, it totally depends upon user. So using jquery "serialize" to get data inserted by user. 

That is not getting me data, as 'name' attribute is missing from inputs. Can you guys please fix it

hansi90:
Add name attribute for input field with slugify in checklist fields. Fix #243

### DIFF
--- a/dev/checklist/app.vue
+++ b/dev/checklist/app.vue
@@ -23,6 +23,7 @@ export default {
 						label: "Skills",
 						model: "skills",
 						required: true,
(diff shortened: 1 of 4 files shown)


icebob:
Thanks, nice job @hansi90!

PR #314, opened 2017-10-11
File upload error fixed.

### LINKED ISSUE #172: File upload

hansi90:
Created a new field. Use schema type: "file". Fix #172


icebob:
@cristijora What do you think?

cristijora:
@icebob @hansi90 That seems to be fine. Might be a little confusing for people to have it separate from input. If that's ok with you my only concern is the naming of the field. Would prefer `fieldUpload`

PR #321, opened 2017-10-17
Set success if the model value is not null

cyaiox:
Set class 'validationSucess' if the model has no errors and its initial value is not null


icebob:
I think it's same as #257. But if model value is not null it doesn't mean that it's touched. For example if you load an existing record the name field won't be null, but doesn't mean it's touched.

cyaiox:
You're right, it's same as #257 but very basic, this feature try to verify when :isNewModel="true" don't set class success to the field to keep neutral state, but if the model is filled, apply the validations and set class success or error to the field.

PR #332, opened 2017-11-03
Fix tag prop

### LINKED ISSUE #331: "tag" not working with grouped schema
Node: version 8.9.0
VGG: version 2.1.1

_**Problem**_
Adding tag="div" still uses fieldset when the schema has grouped fields.

Example:
```vue
<vue-form-generator 
 :model="model" 
 :schema="fields" 
 :options="formOptions" 
 ref="someForm" 
 tag="div" 
 >
 </vue-form-generator>
```

This works with a regular schema.

Ex. (works)
```javascript
schema: {
    fields: [
        {
            type: 'input',
            inputType: 'input',
            label: 'First Name',
        },
        {
            type: 'input',
            inputType: 'input',
            label: 'Last Name',
        },
    ],
}
```

With a grouped schema it does not work.

Ex. (does not work)
```javascript
schema: {
    groups: [
        {
            fields: [
                {
                    type: 'input',
                    inputType: 'input',
                    label: 'First Name',
                },
            ],
        },
        {
            fields: [
                {
                    type: 'input',
                    inputType: 'input',
                    label: 'Last Name',
                },
            ],
        },
    ],
}
```

_**Expected Outcome**_
When adding tag to VFG and having grouped fields, it will change fieldset to the tag specified.

hansi90:
The `tag` prop is fixed in case of groups. Fix #331

### DIFF
--- a/src/formGenerator.vue
+++ b/src/formGenerator.vue
@@ -17,7 +17,7 @@ div.vue-form-generator(v-if='schema != null')
 					span(v-for='(error, index) in fieldErrors(field)', track-by='index') {{ error }}
 
 	template(v-for='group in groups')
-		fieldset
+		fieldset(:is='tag')
 			legend(v-if='group.legend') {{ group.legend }}
 			template(v-for='field in group.fields')
 				.form-group(v-if='fieldVisible(field)', :class='getFieldRowClasses(field)')


icebob:
Thanks! Great job!

Issue #337, opened 2017-11-08
styleClasses for groups, and nested group support


zoul0813:
When groups are processed, it would be nice to be able to set a "class" on the fieldset which would allow for things like Bootstrap's "col-md-6" to have fieldsets side-by-side in a 2-column layout.

Nested groups would also be nice to have with the ability to change the groups "wrapper" element from "fieldset" to a group property.  For example:

```
schema: {
  groups: [
    {
      label: "Name",
      fields: [
        { type: "input", inputType: "text", label: "First Name", model: "first_name" },
        { type: "input", inputType: "text", label: "Last Name", model: "last_name" },
      ]
    },
    {
      label: "Address",
      styleClasses: "row",
      tag: 'div',
      groups: [
        {
          tag: 'div',
          styleClasses: "col-md-6",
          fields: [
            { type: "input", inputType: "text", label: "Street", model: "address1" },
            { type: "input", inputType: "text", label: "Apt/Suite", model: "address2" },
          ]
        },
        {
          tag: 'div',
          styleClasses: "col-md-6",
          fields: [
            { type: "input", inputType: "text", label: "City", model: "city" },
            { type: "input", inputType: "text", label: "State", model: "state" },
          ]
        },
      }
    }
  }
}
```
Which would produce markup similar to this:

```
<fieldset>
  <legend>Name</legend>
  <div class="form-group field-input">...</div>
  <div class="form-group field-input">...</div>
</fieldset>
<div class="row">
  <div class="col-md-6">
    <div class="form-group field-input">...</div>
    <div class="form-group field-input">...</div>
  </div>
  <div class="col-md-6">
    <div class="form-group field-input">...</div>
    <div class="form-group field-input">...</div>
  </div>
</div>
```

zoul0813:
Closing.

PR #339, opened 2017-11-08
Feature/group styles 

zoul0813:
Partial fix for #337 - adds support for "styleClasses" within groups.


icebob:
Travis is red. I think cause of the updated Vue version. Please fix it.

zoul0813:
Sorry about that, was having a version conflict on my end due to other deps in my project and committed that update to the wrong branch.

PR #342, opened 2017-11-10
added labelClasses support

### LINKED ISSUE #333: How to add label icons?
For example I'd like to have a fontawesome icon as label. I could paste this as label icon that I could then style with CSS:

    <i class="fa fa-envelope-o" aria-hidden="true"></i>

But currently it is parsed as text so can't use HTML?

zoul0813:
fixes #333 - adds "labelClasses" to the field schema, allowing you to configure icons and other styles to labels via CSS classes.

### DIFF
--- a/src/formGenerator.vue
+++ b/src/formGenerator.vue
@@ -3,7 +3,7 @@ div.vue-form-generator(v-if='schema != null')
 	fieldset(v-if="schema.fields", :is='tag')
 		template(v-for='field in fields')
 			.form-group(v-if='fieldVisible(field)', :class='getFieldRowClasses(field)')
-				label(v-if="fieldTypeHasLabel(field)", :for="getFieldID(field)")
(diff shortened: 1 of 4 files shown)


zoul0813:
Not too familiar with your testing policies ... does the formGenerator test cover it?

src/formGenerator.vue
```diff
@@ -571,3 +569,11 @@ div.vue-form-generator(v-if='schema != null')
 
 	} // fieldset
 </style>
+ // .errors
```
zoul0813: oops ... how'd that get there?

PushpinderRattan:
labelClasses field does not seem to be working for me as I am trying to add a fontawesome icon next to the field. Is there anyway to do so? Do you have any example?
Thank you.

PushpinderRattan:
Please ignore it.

PR #350, opened 2017-11-16
Feature/347 no empty labels

### LINKED ISSUE #347: don't render labels when no label text is provided
While working on custom checkbox components that include a label, I've discovered that all fields have a label tag rendered when "fieldTypeHasLabel" returns true.  I propose adding an additional check to this function to return false when the field's schema does not have a label property (or altering the v-if statement on the label.

```
// option 1: add the following to fieldTypeHasLabel(field)
if(isNil(field.label)) 
  return false;
```

```
// option 2: alter the label v-if
label(v-if="field.label && fieldTypeHasLabel(field)", :for="getFieldID(field)", :class="field.labelClasses")
```

This requires the unit tests to be aware of this change as well.  

I can submit a PR for this change if approved.  Just need to know which option makes more sense.

zoul0813:
resolves #347

### DIFF
--- a/src/formGenerator.vue
+++ b/src/formGenerator.vue
@@ -205,6 +205,8 @@ div.vue-form-generator(v-if='schema != null')
 
 			// Should field type have a label?
 			fieldTypeHasLabel(field) {
+				if(isNil(field.label)) return false;
+
 				let relevantType = "";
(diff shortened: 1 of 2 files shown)


icebob:
Sorry, I forgot it.

PR #359, opened 2017-12-12
Include UID for emitted form validation events

patcarter883:
Added a UID to from validation events to allow tracking validation of forms inside of a vue template loop.


src/formGenerator.vue
```diff
@@ -307,7 +307,7 @@ div.vue-form-generator(v-if='schema != null')
 				}
 
 				let isValid = this.errors.length == 0;
-				this.$emit("validated", isValid, this.errors);
+				this.$emit("validated", isValid, this.errors, this._uid);
```
zoul0813: It looks like you're relying on the internal _uid from the Vue component, this property is undocumented and meant for internal use by Vue.  If you'd like to implement something like this, then I'd recommend you introduce a stable property and initialize it with a value we can rely on.

zoul0813:
Maybe it would make more sense if we returned the VFG instance in the validated event instead of a UID or tracking number.  This would allow developers to have access to the entire VFG instance .... 

Something like this?
```
this.$emit('validated', isValid, this.errors, this);
```

Devs could then use the internal _uid property, or add additional attributes to the VFG tag ... such as `:id="someGeneratedId` and then access that, they'd also have access to the model and schema in the VFG instance to respond to `validated` events properly.

patcarter883:
Sounds like a plan. Done!!

PR #360, opened 2017-12-12
fixes #341 - introduced debounce functionality into `formatValueToModel`

### LINKED ISSUE #341: fieldInput - invalid date in fecha.format
When entering dates manually using the keyboard in Chrome, fecha throws errors.

JSFiddle: https://jsfiddle.net/zoul0813/21ba3on5/

```
{
    type: "input",
    inputType: "date",
    label: "Date",
    model: "date_field",
    format: "YYYY-MM-DD"
}
```

When the input is focused, you can enter the date via the keyboard, such as "01", "01", "1980".  As soon as you type the "1" for the year fecha throws an error because the Date field currently has "01/01/0001" but Chrome allows you to continue appending to the value by entering the other 3 digits. fecha throws 3 errors during this time, as "0001", "0019" and "0198" are invalid years but the user has not finished entering the date.

The same error occurs when you use the "x" in Chrome to clear the field. When clearing the field value, the value is not "null" and appears to be an empty string.  fieldInput checks against null.

Once the date has been made valid, the model is not properly updated to reflect the correct value. Once the fecha errors are thrown, data binding appears to break ... clearing the field with the "x" and using the built-in date dropdown prevents the model from being updated as well.

zoul0813:
* Updated to fecha 2.3.2

* Moved the majority of logic from computed.value.set() to new `updateModelValue(newValue, oldValue)` method to allow other components to use similar techniques.
If `formatValueToModel` returns a function, then the function is called without setting the model value.  It is the functions responsibility to call `this.updateModelValue(newValue, oldValue)`.
Otherwise, calls `this.updateModelValue(newValue, oldValue)` to execute the original logic.

* If the input type is date, datetime, or datetime-local then it creates a debounced function for setting the value in fieldInput's mounted hook and then returns a function that calls this debounced function from `formatValueToModel`

* Reduced the repeated code for date formatting by adding a DATETIME_FORMATS constant, and using that to
get the "format" for fecha.parse()


icebob:
Hmm, Travis is not working after organization. I'm fixing.

icebob:
Could you modify a file (some space or enter) that the travis will be triggered?

zoul0813:
Well that's interesting, `npm run test` and `npm run build` ran without error on my end ... is Travis using different settings?

icebob:
Travis was linked to the old "icebob/vue-form-generator" repo :)
I changed it to the new organization repo, now it's working.

zoul0813:
I was referring to the failure - Travis said I missed a comma and failed the validation, but on my end everything passed without error/warnings.

icebob:
Hmm, interesting.

PR #363, opened 2017-12-13
fixes #361 - use $event.target.valueAsNumber for number/range inputs,…

### LINKED ISSUE #361: Value of number type input
I saw the way of getting value of a number type input is using "this.value".
And problems happened:
when the user enter  like this "10e10",this will be OK
But when they continue to type just like this "10e10ee.12e12e12e.12e21e".Obviously,this input is not a valid number,but the "this.value" equals to the Empty string,And looks like you guys converted the string to number.So the empty string will be converted to 0. And if we set a negative min/max value,the “10e10ee.12e12e12e.12e21e” will be passed, because 0 is greater than negative value.This is not what we expect.So maybe you can use the "this.valueAsNumber".In this way,the "10e10ee.12e12e12e.12e21e" will be pasred to NaN .And we can handle the error condition more  clearly .

zoul0813:
fixes #361 - use $event.target.valueAsNumber for number/range inputs, debounce `formatValueToModel` for number/range, removed `formatValueToField`

Number inputs now support `10e10` formats being typed in, and stores the evaluated value of `10e10` (`100000000000`) on the model.

* moved debouncedFormatDateTime to abstractField and renamed to debounceFormatFunc
* create a `debounceFormatFunc` for number/range inputs, as well as date/datetime/datetime-local
* changed @input to use `onInput` handler which checks for number/range and uses $event.target.valueAsNumber per @YgamiLight
* set value to NaN if !isNumber(value) (NOTE: NaN is represented as "null" in JSON though as JSON does not support NaN/Infinity)
* set all `inputType` checks and uses to use `inputType.toLowerCase()`, recommend doing this for all fields
* created DEBOUNCE_FORMAT_MS constants in inputField.vue, set to default of 1000 ... recommend turning this into a standard schema option for all fields
* wrapped one-liner conditionals with `{}` (just looks cleaner)


icebob:
@zoul0813 You can merge it if it's done.

lionel-bijaoui:
In IE11, `valueAsNumber` return `NaN` for any value.
In Edge, there is also problem with it.
https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/669685/#error

I'm going to make a PR

YgamiLight:
@lionel-bijaoui  Thanks, you really did a great job!

PR #365, opened 2017-12-13
fixes #340 - "none" value set to `null`, formatValueToField checks fo…

### LINKED ISSUE #340: fieldSelect.vue - noneSelected is not visible, required is not set
The "none selected" option is not "selected" by default, so the drop down appears empty.

Here's a JSFiddle, https://jsfiddle.net/zoul0813/ooncsad3/

The expectation is for the "Select One" text to appear in the dropdown if the user has not modified the dropdown value yet, the same way the "placeholder" attribute on `input` elements work. `:selected="value == undefined"` appears to not be working ... even though value is undefined.

fieldSelect.vue also does not set the "required" attribute on the `select` when `schema.required` is true.

zoul0813:
fixes #340 - "none" value set to `null`, formatValueToField checks for `isNil(value)` and returns `null`, none options are always disabled

* updated tests to reflect "none is always disabled", could possibly be a selectOptions to allow users to select the "none" option to clear a previously selected value?
* added formatValueToField and return `null` when `isNil(value)`


src/fields/core/fieldSelect.vue
```diff
@@ -1,6 +1,6 @@
 <template lang="pug">
 	select.form-control(v-model="value", :disabled="disabled", :name="schema.inputName", :id="getFieldID(schema)", :class="schema.fieldClasses")
-		option(v-if="!selectOptions.hideNoneSelectedText", :disabled="schema.required", :value="null", :selected="value == undefined") {{ selectOptions.noneSelectedText || "&lt;Nothing selected&gt;" }}
+		option(v-if="!selectOptions.hideNoneSelectedText", :disabled="true", :value="null") {{ selectOptions.noneSelectedText || "&lt;Nothing selected&gt;" }}
```
icebob: I think the `schema.required` was a good solution, because `null` can be a good & valid value.

zoul0813: reverting this change back to `schema.required`

zoul0813:
Merged with `master`, bringing in changes from PR #363  and #364

icebob:
Thanks. You can merge it.

PR #366, opened 2017-12-15
fixes #358 - support "validateBeforeSubmit" with async validators

### LINKED ISSUE #358: Asynchronous Validators are broken
I found two major breaks in the asynchronous validation. I reproduced both in this [jsfiddle](https://jsfiddle.net/0mg1v81e/964/)

First, if I type multiple characters into a text field, for each typed character a new validation error is displayed. Even if using `formOptions.validateDebounceTime`, which is currently not documented, multiple validation errors are displayed if I type a new character after the debounce timer and before the asynchronous validation is completed.

Second, asynchronous validation errors are not included when using a submit button with the following schema:
```
{
    type: 'submit',
    validateBeforeSubmit: true,
}
```
I can still submit the form even if asynchronous validation errors are currently displayed.

zoul0813:
fixes #358 - support "validateBeforeSubmit" with async validators

* added `validateAsync` (default: false) form option
* added `isAsync` parameter to FormGenerator.validate()
* FormGenerator.validate() returns Promise when `validateAsync` or `isAsync` is true
* renamed `click` handler to `onClick` in FieldSubmit
* added `onValidationError(model, schema, errors)` to FieldSubmit schema to handle validation errors
* added async validator support for FieldSupport
* changed `each` to `forEach` in various places, as "each" is an alias to "forEach" and "forEach" looks more explicit
* removed call to Vue.util.hyphenate as this is no longer supported by Vue, replaced with equivalent `String.replace` expression
* updated fieldSubmit.spec to add "valid form" and "invalid form" tests, valid forms will always call `onSubmit`, invalid forms will not call `onSubmit` (when validateBeforeSubmit = true)
* various code clean up

To use the new async validation logic, just set `validateAsync: true` in your form options, or call FormGenerator.validate(true) manually.

Code should be backwards compatible.  Additional tests are likely needed to test the new async logic.

### DIFF
--- a/src/fields/abstractField.js
+++ b/src/fields/abstractField.js
@@ -59,7 +59,6 @@ export default {
 
 	methods: {
 		validate(calledParent) {
-			// console.log('abstractField', 'validate', calledParent);
 			this.clearValidationErrors();
 			let validateAsync = objGet(this.formOptions, "validateAsync", false);
 
@@ -96,17 +95,14 @@ export default {
 			}
 
 			let handleErrors = (errors) => {
-				// console.log('abstractField', 'all', errors);
 				let fieldErrors = [];
 				forEach(errors, (err) => {
-					// console.log('abstractField', 'err', err);
 					if(isArray(err) && err.length > 0) {
 						fieldErrors = fieldErrors.concat(err);
 					} else if(isString(err)) {
 						fieldErrors.push(err);
 					}
 				});
-				// console.log('abstractField',  'fieldErrors', 'final', fieldErrors);
 				if (isFunction(this.schema.onValidated)) {
 					this.schema.onValidated.call(this, this.model, fieldErrors, this.schema);
 				}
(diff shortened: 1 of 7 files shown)


src/fields/abstractField.js
```diff
@@ -60,52 +60,66 @@ export default {
 	methods: {
 		validate(calledParent) {
 			this.clearValidationErrors();
+			let validateAsync = objGet(this.formOptions, "validateAsync", false);
+
```
icebob: Why need the `validateAsync` option? Is it not enough if we check the type of result of `validator` function (Promise or not)?

zoul0813: The updated code returns a promise if validateAsync is true, and waits for async validation to complete ... this is not what the current code does, so the new option is a way to enable it while maintaining backwards compatibility so we don’t break forms that expect the validate function to return a Boolean.

zoul0813: V3 would likely remove this option and always return a promise from validate... making it the default validation mode, which is what most people are likely expecting from the current async validation.

icebob: Ok. in v3 we change it to easier (handle all as Promise)

PR #370, opened 2018-01-09
Unscaped html for error msg

gezichenshan:
Adding a schema property named errorUnescaped to fulfill the situation that the severside return a string with unescaped html.
see: https://jsfiddle.net/7of916d4/2/


icebob:
Is it important to create a new field property? Can we use always `v-html` to show the error messages?

gezichenshan:
I think you are right. But I've no idea whether this always v-html will result in some XSS attack?

icebob:
I think XSS is not relevant here because we don't render user input texts, just error messages what comes from the developer. What do you think?

gezichenshan:
Yep, I agree with you. So I'll remove the field property and add `v-html` on the error part.

icebob:
Ok :+1:

src/formGenerator.vue
```diff
@@ -14,8 +14,7 @@ div.vue-form-generator(v-if='schema != null')
 						button(v-for='btn in field.buttons', @click='buttonClickHandler(btn, field, $event)', :class='btn.classes') {{ btn.label }}
 				.hint(v-if='field.hint') {{ fieldHint(field) }}
 				.errors.help-block(v-if='fieldErrors(field).length > 0')
-					span(v-for='(error, index) in fieldErrors(field)', v-if='field.errorUnescaped', v-html="error", track-by='index')
-					span(v-for='(error, index) in fieldErrors(field)', v-if='!field.errorUnescaped', track-by='index') {{ error }}
+				span(v-for='(error, index) in fieldErrors(field)', v-html="error", track-by='index')
```
icebob: Add indent

icebob:
Thanks!

gezichenshan:
With pleasure.

PR #410, opened 2018-03-07
fixes #409 - added ":" to id property in template

### LINKED ISSUE #409: fieldUpload not setting id attribute correctly
Setting the `id` property in the schema for field `type: "upload"` results in markup that looks like:
```
<input id="getFieldID(schema)" 
     type="file"
     accept=".jpg,.jpeg,.png,.pdf" 
     required="required" class="form-control">
```

Comparing it to other core field files, it looks like in `fieldUpload.vue` the `id` prop is missing a colon, [here](https://github.com/vue-generators/vue-form-generator/blob/master/src/fields/core/fieldUpload.vue#L4).

Could be an easy/quick fix :)

zoul0813:


icebob:
Should I release a new version?

zoul0813:
a minor release perhaps - is there a way to automate those?

icebob:
I'm going to try to setup automate releasing.

icebob:
@zoul0813 https://github.com/vue-generators/vue-form-generator/issues/246#issuecomment-371791641

Issue #425, opened 2018-03-21
Number fields delete '0' and '-' on debounce


kretep:
* **I'm submitting a ...**
  * [x] bug report

* **What is the current behavior?**
Number fields don't handle '0' and '-' in some cases; they are deleted at debounce. To reproduce:
  * https://jsfiddle.net/0mg1v81e/1978/
  * Make sure a valid number > 0 is in the input box
  * Select all text in the input box
  * Type a 0 or - and wait for debounce
  * On debounce, the input box will be emptied

* **More details and clues**
  * When typing a 0 or - for a second time, the value will stay
  * It only seems to happen when you type the 0 or - when all text is selected and when a number > 0 was previously in the box.
  * In the method [formatNumberToModel](https://github.com/vue-generators/vue-form-generator/blob/master/src/fields/core/fieldInput.vue#L87), the newValue is sometimes a number and sometimes a string; I don't understand why.
  * When replacing the isNumber() check with parseFloat, this solves some of the problems.

* **What is the expected behavior?**
  I expect '0' and '-' not to be deleted when entered in a text field

* **Please tell us about your environment:**
  * Version: 2.2.1
  * Browser: Chrome

andrelec1:
Well i have same issue ....
i try to fixe it but ....

![image](https://user-images.githubusercontent.com/3273204/38314067-bf1f3612-3825-11e8-9b1a-db69de2905c4.png)

zoul0813:
‘0’ is number like, but it’s technically a string.  I don’t agree that ‘0’ should be considered a number.  This seems like an issue with the data model as it is storing the value as a string instead of a number.  The input field should be providing the number and not a string representation of the number.

The negative sign should be treated properly though, and we’ll have to address that.  Perhaps by validating it as true while the field has focus, and false when the field loses focus (which would imply the user only typed - into the field which is invalid)

kretep:
This indeed resolves the issue for the zero, but not for the minus sign.

I kind of like your suggestion of handling the validation differently when the field has the focus.

zoul0813:
Looking into it, it's quite complicated to test if the input field has only a "-" in it or not ... the only way I've managed to get a negative into the field is by racing the "debounce" clock which is set to 1s ... we can probably allow this value to be set manually.

[...]

PR #430, opened 2018-04-04
Fix NaN with value 0 on input type number/range.

andrelec1:

Add simple *1 on value for forcing number type.

https://github.com/vue-generators/vue-form-generator/issues/425

Bug fix

No

* **Other information**:


zoul0813:
I don't agree with these changes - but am open to being convinced otherwise as I may be missing something here.

NaN is the appropriate value to be used when the user enters anything that is "not a number".  I also don't agree that numbers stored as strings '0', '1', '2', etc ... should be treated as anything other than strings.  The field expects a number, `parseInt('0')` or `new Number('5')` should be used to convert string values into their number representation before passing the model to VFG.

andrelec1:
I found the real issue, and change my commit ...

zoul0813:
What’s wrong with the NaN result? Is the browser having an issue displaying NaN in the input?

andrelec1:
in fact, if you put 0 if the field, you didn't convert the value in number  ... ( if (0)  => false )

```onInput($event) {
			let value = $event.target.value;
			switch(this.schema.inputType.toLowerCase()) {
				case "number":
				case "range":
					if($event.target.valueAsNumber) {
						value = $event.target.valueAsNumber;
					}
					break;
			}
			this.value = value;
		},
```
So when value go throut `formatNumberToModel` ( my first wrong fix :/ ) you have a string ... so you put NaN in model ...
In some case you realy want put 0 in field and get 0 in model ... 0 is valide number ... 

https://jsfiddle.net/0mg1v81e/1978/

qkdreyer:
I can confirm this fixes #425.
You can check it here : https://jsfiddle.net/0mg1v81e/2323/

zoul0813:
Not sure I understand what's being fixed here.

Here's the same JSFiddle, pointing to the official 2.1.1 version of VFG ... same result, unless I'm missing something?

I'm able to type in -10, -25, 0, etc ... and the form validates them all, what am I missing here?

qkdreyer:
Without PR : http://recordit.co/bjcamzQq62
With PR : http://recordit.co/XvuAQjTNBV

zoul0813:
Do you get the same result when using my fiddle against 2.1.1?

qkdreyer:
I'm sorry but I am not able to find your fiddle...

zoul0813:
Doh, I didn't link it ... https://jsfiddle.net/94dcctbf/3/

qkdreyer:
Could you hotfix/merge this ASAP since it’s a BC minor change and some of us rely on this library in production ?

PR #435, opened 2018-04-10
Add npm script and docs to run examples

slathrop:

* **What kind of change does this PR introduce?**
Documentation and convenience npm script

- **What is the current behavior?**
Running the examples is not obvious

Running the examples is now documented and obvious

- **Does this PR introduce a breaking change?**
No


zoul0813:
The “examples” script appears to do the same as the dev script, except it opens the browser? Should we not just modify “dev” to open the browser instead? Those aren’t really examples, they’re test pages for the various components.

A proper example would maybe be a github hosted page showing the various components in action with live feedback, and maybe some stylized code?

slathrop:
Hi @zoul0813, the `examples` script does a few things differently:

1. Sets `--content-base` to the root instead of the `dev` folder, making it *possible* to navigate to `examples`
1. Uses `--no-info` to hide the thousands of lines of output seen when using `npm run dev`
1. And yes, opens the browser

Perhaps make your suggestion for a "proper" example as a new issue?

This PR is just an attempt by someone coming to this repo for the very first time this week and trying to get some value out of it and give back a little value as well, FWIW.

zoul0813:
@slathrop unfortunately though, I don't see this update providing much benefit to people who are not actively developing for VFG.  The examples folder is self-running by just opening the files in the browser such as `examples/simple/index.html` and does not require the webpack dev server to run.

The other problem is that the new script does not work unless you checkout the github code, or install VFG in dev mode, which is not the normal process unless you're writing code for VFG itself.

I appreciate the effort, but I don't think this PR serves much use as it just adds an extra step for people who wish to just implement the library.  At the moment, they can just double click the example html files and open them in a browser locally without a server.

If you'd like to work on better examples, and provide those in a way we can host on github pages - that would be great.  We can also maybe keep the update to the README.md if it's rewritten to explain the examples are found in the `examples/` folder and can be opened directly from there (if they download the full VFG source).

Issue #408, opened 2018-03-07
Add event in callback onValidationError


zevitagem:
It is necessary to add the $ event variable in the onValidationError callback in of the "FieldSubmit.vue" file.

In my case, this component is inside a FORM, and the way the code is currently, I can not do "preventDefault" when the form has errors, I can only do it on the "onSubmit".

zoul0813:
@zevitagem I recommend not placing VFG inside of a <form /> ... but if you do, you should attach a submit handler to the form and check VFG's validation status in there ...

```
<form @submit="onSubmit">
  <vue-form-generator refs="vfg" ... />
</form>
```

```
onSubmit($event) {
  if(this.$refs.vfg.errors.length > 0) {
    $event.preventDefault();
    // handle invalid form
  }
}
```

Alternatively, you can just use `@submit.prevent="onSubmit"` and then submit the form manually in the handler.

jaywilburn:
I am using VFG but from generated JSON. I parse out the function and write it outside the JSON. How would I set preventDefault? Right now this is being used in a Vue app that lives inside of an ASP.net site, and the submit causes a post back. How can I write the function to submit but also preventDefault?

zoul0813:
@jaywilburn attach an event callback to the forms submit event, and catch it there.  Do not attempt to catch the submit from the button's event.

tim-yao:
Just would like to correct a typo in above solution https://github.com/vue-generators/vue-form-generator/issues/408#issuecomment-371148616

Should be `ref` not `refs`.

```<form @submit="onSubmit">
  <vue-form-generator ref="vfg" ... />
</form>```

zoul0813:
Thanks, I fixed the comment

PR #443, opened 2018-04-22
closes #434, #408 - added $event to onValidationError

### LINKED ISSUE #434: validateBeforeSubmit doesn't prevent the submit
### Current : 
When an input submit has `validateBeforeSubmit` setted to `true` the form is not prevented from submitting and call the `onSubmit` callback when there is errors.

### Expected :
It should not submit if there is errors.

zoul0813:
added $event to onValidationError and call preventDefault when onValidateBeforeSubmit is true

We can't prevent the submit after waiting for async validation, so we just prevent it before calling any validation and let the user
manually submit the form in the `onSubmit` handler, if it validates.

Bug Fix

`<form />` are submitted when using `submit` field with validateBeforeSubmit.

`validateBeforeSubmit: true` calls `$event.preventDefault()` and prevents the form submission 

This PR should introduce "expected behavior", so users should not need to do anything.

* **Other information**:

### DIFF
--- a/src/fields/core/fieldSubmit.vue
+++ b/src/fields/core/fieldSubmit.vue
@@ -12,10 +12,13 @@
 		methods: {
 			onClick($event) {
 				if (this.schema.validateBeforeSubmit === true) {
+					// prevent a <form /> from having it's submit event triggered
+					// when we have to validate data first
+					$event.preventDefault();
 					let errors = this.$parent.validate();
 					let handleErrors = (errors) => {
 						if(!isEmpty(errors) && isFunction(this.schema.onValidationError)) {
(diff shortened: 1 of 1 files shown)


zoul0813:
@icebob I'm going to merge this in later today as no one else has reviewed it. OK?

Also - if I bump the version that will automatically do an npmjs update?

icebob:
OK.

Yes, but you have to use `npm run version ....` script. More info: https://github.com/vue-generators/vue-form-generator/issues/246#issuecomment-371791641

zoul0813:
@icebob yes, I'm going to do that a separate PR with the version bump and include refs to the issues included in it.  Thanks.

PR #467, opened 2018-06-20
Rework and html for label & hint

lionel-bijaoui:

Feature: Allow for HTML in the label and hint field

Text only for label and hint
Full html for label and hint
No
* **Other information**:
### *Build on top of the v3 branch*
This work is following the changes introduced in my latest PR and should be merged only after that.
If my PR #465 is rejected, I will cherry-pick the necessary commit, deal with the conflicts and remake this PR. This will take me time for nothing, so I hope #465 will be accepted and merged.
I did that to simplify the work for the v3 features and rework.
By introducing some of these changes right now, I hope to prepare the code for the future.

### New component
Part of `formGenerator.vue` was put inside another component. The necessary changes were made and the test changed to fit the new structure.
It should be almost transparent to the user, except if they used complicated way to access the fields from `<vue-form-components>` (eg. with ref like `this.$refs.vfg.children[0]`). And even then, I tried to make it as transparent as possible.

### Doing without Pug
As you may notice, this PR is not using Pug for the new component.
I propose we stop using Pug in the future and that we slowly remove it.
Here are the reason why :
- ESlint doesn't work in Pug template, so we get no lintting from eslint-plugin-vue
- Pug can be a barrier to entry to collaborator as they need to learn (or know) this specific syntax to help
- Pug is not bringing new functionnlity/solution (vue template are dynamic enough and Emmet is a thing)
- To use Pug, we need additionnal plugin/parser/linter to be installed (can introduce problem/incompatibility)


src/fields/optional/fieldNoUiSlider.vue
```diff
@@ -1,116 +1,117 @@
 <template>
-	<div class="slider" :disabled="disabled" :class="{ 'contain-pips': containPips, 'contain-tooltip': containTooltip }"></div>  
+	<div class="slider" :disabled="disabled" :class="{ 'contain-pips': containPips, 'contain-tooltip': containTooltip }"></div>
```
dflock: This not getting `v-attributes`?

src/utils/schema.js
```diff
@@ -54,45 +51,52 @@ module.exports.mergeMultiObjectFields = function (schema, objs) {
...
+const slugify = (name = "") => {
+	// Return the slugified version of either:
+	return (
+		name
 			// NB: This is a very simple, conservative, slugify function,
 			// avoiding extra dependencies.
 			.toString()
 			.trim()
-			.toLowerCase()
+			// .toLowerCase()
```
dflock: Why comment this line out?

src/utils/schema.js
```diff
@@ -54,45 +51,52 @@ module.exports.mergeMultiObjectFields = function (schema, objs) {
...
 			// Spaces & underscores to dashes
-			.replace(/ |_/g, "-")
+			.replace(/ /g, "-")
```
dflock: If you're going to change the output & meaning, change the comments to match.

src/utils/schema.js
```diff
@@ -54,45 +51,52 @@ module.exports.mergeMultiObjectFields = function (schema, objs) {
...
 			// Multiple dashes to one
 			.replace(/-{2,}/g, "-")
 			// Remove leading & trailing dashes
 			.replace(/^-+|-+$/g, "")
 			// Remove anything that isn't a (English/ASCII) letter, number or dash.
```
dflock: If you're going to change the output & meaning, change the comments to match.

src/utils/schema.js
```diff
@@ -54,45 +51,52 @@ module.exports.mergeMultiObjectFields = function (schema, objs) {
...
 	// Try to get a reasonable default id from the schema,
 	// then slugify it.
 	if (typeof schema.id !== "undefined") {
 		// If an ID's been explicitly set, use it unchanged
 		return prefix + schema.id;
 	} else {
 		// Return the slugified version of either:
-		return prefix + (schema.inputName || schema.label || schema.model || "")
+		return (
+			prefix +
```
dflock: Are you duplicating the `slugify` function in here?

dev/mixinUtils.js
```diff
@@ -1,12 +1,19 @@
-module.exports = {
-	
-	filters: {
-		prettyJSON: function(json) {
+export default {
```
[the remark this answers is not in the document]
lionel-bijaoui: I'm changing that no problem

package.json
```diff
@@ -1,113 +1,151 @@
...
+	"browserslist": [
+		"last 2 versions"
+	],
+	"prettier": {
+		"printWidth": 120,
+		"tabWidth": 4,
+		"singleQuote": false,
+		"trailingComma": "none",
+		"bracketSpacing": true,
+		"semi": true,
```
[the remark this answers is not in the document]
lionel-bijaoui: Why remove the options ?

src/fields/optional/fieldNoUiSlider.vue
lionel-bijaoui: I didn't think about it. Since fields will be separated from the main project (in the v3), I'm not touching them too much for now.

src/utils/schema.js
lionel-bijaoui: TBH, I have no idea... :/
I'm going to revert that

lionel-bijaoui: It come from that https://github.com/vue-generators/vue-form-generator/commit/17aba74

src/utils/schema.js
lionel-bijaoui: Again, it come from that  https://github.com/vue-generators/vue-form-generator/commit/17aba74
I'm going to make the change you suggested

src/utils/schema.js
lionel-bijaoui: Again, it come from that 17aba74
I'm going to make the change you suggested

package.json
```diff
@@ -1,113 +1,151 @@
...
-    "webpack-merge": "0.14.1"
-  },
-  "dependencies": {},
-  "publishConfig": {
-    "access": "public"
-  }
+	"name": "vue-form-generator",
+	"version": "3.0.0-dev",
+	"description": "A schema-based form generator component for Vue.js",
+	"homepage": "https://github.com/icebob/vue-form-generator",
```
zoul0813: https://github.com/vue-generators/vue-form-generator/

src/utils/schema.js
```diff
@@ -54,45 +51,52 @@ module.exports.mergeMultiObjectFields = function (schema, objs) {
...
+			(schema.inputName || schema.label || schema.model || "")
+				// NB: This is a very simple, conservative, slugify function,
```
zoul0813: we already rely on lodash, why not just use _.kebabCase() here so it's a bit clearer and shorter?

src/utils/schema.js
lionel-bijaoui: Again, it come from 17aba74, I have no idea why there is two slightly different slugify...

src/utils/schema.js
lionel-bijaoui: Again, I'm sorry, I have no idea why we do that. This is a good suggestion but I'm not going to implement too much changes in this PR (there is already a lot of changes). Maybe next PR ? Or v3 ?

package.json
lionel-bijaoui: Yes, sorry, I'm going to fix that

lionel-bijaoui:
@dflock @zoul0813 Thank you for your feedback. Is it better now ?

src/utils/schema.js
zoul0813: Sounds good - we should probably do a full code review at some point and simplify whatever we can, relying on known functions with clear documentation (such as kebabCase).

zoul0813:
This is backwards compatible with v2 correct? Should the version be 2.2.3 or 2.3.0 ... instead of 3.0.0-dev?  Sorry, missed that during the initial review.

package.json
```diff
@@ -2,7 +2,7 @@
 	"name": "vue-form-generator",
 	"version": "3.0.0-dev",
```
zoul0813: should be 2.2.3 or 2.30 - I'm thinking 2.3.0 due to all the changes?

src/utils/schema.js
lionel-bijaoui: You are absolutely right !

package.json
lionel-bijaoui: Yes, sorry. It was the v3 branch initially but I actually only made backward compatible changes. So I will call it 2.3.0, it will require `npm i` for all the collaborator since the test and plugin are updated, but it will change nothing for the end user. I tested it on my project without any problem and I use fairly complex custom fields, an api to generate the schema, complex dynamic functionality and they needed no changes and introduce no bug.
I might have missed something of course, but I'm pretty confident.

lionel-bijaoui:
@dflock That should be good

package.json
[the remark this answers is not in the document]
dflock: Personally prefer less indentation and less semicolons `¯\_(ツ)_/¯`

src/fields/optional/fieldNoUiSlider.vue
dflock: OK

package.json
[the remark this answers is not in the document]
lionel-bijaoui: With Prettier, everything is done automatically. Even if you write your code a certain way, Prettier format it to this way on save (add semicolons, fix indentation). You don't have to change your habits and it allow for an unified way of writing.
I based myself on the existing code. The indentation is arbitrary, but I have a strong opinion about semicolons. A missing semicolons can cause a bug in some cases, too much semicolons cannot. Also, while parser can understand js without semi, it is not the recommended way to write js (https://stackoverflow.com/questions/444080/do-you-recommend-using-semicolons-after-every-statement-in-javascript).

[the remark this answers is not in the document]
zoul0813: I’d rather there be too many semis than not enough.  /opinion

PR #476, opened 2018-07-02
Fix fields exposition (temporary solution)

lionel-bijaoui:

Fix the fact that, fields are not exposed
You cannot import fields like that `VueFormGenerator.components['fieldInput']`
You should be able to do it like before

* **Other information**:

### DIFF
--- a/src/formGenerator.vue
+++ b/src/formGenerator.vue
@@ -16,6 +16,24 @@ import { get as objGet, forEach, isFunction, isNil, isArray } from "lodash";
 import formMixin from "./formMixin.js";
 import formGroup from "./formGroup.vue";
 
+let fieldComponents = { formGroup };
+
+let coreFields = require.context("./fields/core", false, /^\.\/field([\w-_]+)\.vue$/);
+
+forEach(coreFields.keys(), key => {
+	let compName = key.replace(/^\.\//, "").replace(/\.vue/, "");
+	fieldComponents[compName] = coreFields(key).default;
+});
+
+if (process.env.FULL_BUNDLE) {
+	let Fields = require.context("./fields/optional", false, /^\.\/field([\w-_]+)\.vue$/);
+
+	forEach(Fields.keys(), key => {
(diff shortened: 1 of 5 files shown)


src/formGenerator.vue
```diff
@@ -16,6 +16,24 @@ import { get as objGet, forEach, isFunction, isNil, isArray } from "lodash";
...
+
+	forEach(Fields.keys(), key => {
+		let compName = key.replace(/^\.\//, "").replace(/\.vue/, "");
+		fieldComponents[compName] = Fields(key).default;
+	});
+}
+
 export default {
 	name: "formGenerator",
 	components: { formGroup },
```
zoul0813: you're not actually exposing them in the FormGenerator.components property.  Shouldn't `components: { formGroup }` be `components: { fieldComponents }`?

lionel-bijaoui: Yikes sorry !

lionel-bijaoui: Well, this break a lot of things :(

lionel-bijaoui: @zoul0813 I tried an alternative way to provide the same functionality.
It export `fieldComponents`, so you should be able to do that : 
```js
import { fieldComponents as VueFormGenerator } from 'vue-form-generator'; 
/* ... */
mixins: [ VueFormGenerator['fieldInput'] ],
```
or either
```js
import { fieldComponents as coreFields} from 'vue-form-generator'; 
/* ... */
mixins: [ coreFields['fieldInput'] ],
```

zoul0813: "I" don't need this fix ... but if anyone did the same thing :)

lionel-bijaoui: This should work, have you tried it ?

zoul0813: haven't checked it, no ... but it looks like it will work just fine.

Should formGroup import the fieldComponents from index ... rather than repeating the same logic?

lionel-bijaoui: I have refactored the code

Issue #259, opened 2017-07-17
Group are not flexible enough


lionel-bijaoui:
Hello @icebob @jmverges @dflock 
I'm late to the party and I didn't really follow the work on groups.
Since working on #257 , I'm exposed to how group work and I have noticed a flaw in the current design.
I can't alternate freely between single field and group.
The only possibility is `fields -> groups`. I can't do the inverse (ex. `groups -> fields`) or anything else (ex. `field -> group -> field -> group -> group -> field`)
My goal is not to spit on everyone hard work, but I wonder if they can be a better more flexible way to do this.
Why a `type: group` was not used ?
```js
fields: [{
    type: "input",
    inputType: "text",
    label: "Single field (without group)",
    model: "single"
}, {
    type: "group",
    legend: "Contact Details",
    fields: [{
        type: "input",
        inputType: "text",
        label: "Name",
        model: "name"
    }, {
        type: "input",
        inputType: "email",
        label: "Email",
        model: "email"
    }]
}, {
    type: "group",
    legend: "Other Details",
    fields: [{
        type: "input",
        inputType: "text",
        label: "More",
        model: "others.more"
    }, {
        type: "input",
        inputType: "text",
        label: "Things",
        model: "others.things"
    }]
}]
```
Thanks for the feedback.

icebob:
The current solution was the easiest way to implement it. 
If you use groups as a type of field it is more complicated to implement in the `template`. Need to handle multi-level groups too.

jamestowers:
This would be a nice feature, you could potentially set the group tag and className so you're confined to `fieldset`, eg:

```
type: "group",
legend: "User details",
tag: "div",
className: "col-6 md-col-12",
fields: [{...}]
```

cincarnato:
I really need this
group
{
className: "col-6 md-col-12",
fields: [..

zoul0813:
PR #339 adds "styleClasses" to groups.  This allows you to add a "styleClasses" property to your groups, the same way you add it for fields.  Coupled with the "tag" property, you can now have groups defined as `<div class="col-md-6"></div>`, for example.

Sample Use
```
<vue-form-generator :schema="schema" :model="model" :options="formOptions" tag="div"></vue-form-generator>
```

Sample Schema
```
{
    legend: "Group Title",
    styleClasses: "group-class",
    fields: [
        {
            type: "input",
            inputType: "text",
            label: "Text Label",
            model: "text_field",
            styleClasses: "style-class",
[...]

PR #484, opened 2018-08-27
Better groups and new validation mechanic 

lionel-bijaoui:

New group system and revamped validation system.

#259 
See #259 
Change schema for groups
* **Other information**:
This is still a WIP, I tried to modify the group first but faced many issue related to the current validation system.
So I went back and started to work on validation. Instead of using strongly bonded parent/child relationship, it now use an event bus. All event circulate on this event bus.
As of now, it is more of a POC than the final implementation.
I use the internal property "_uid" of vue objects ~~and I don't think you can use more than one vue-form-generator by page (they will use the same event bus and no id is created, so validation one vfg will trigger all the validation)~~ EDIT: fixed.
I'm going to tackle these issues first, then, implement groups. At this point, validation should be independent from structure and allow more flexible solutions.
@zoul0813  Please take a look if you have time and help me find unforeseen consequences.


lionel-bijaoui:
Now instead of a global event bus for the whole component, a new one is created per instance and send down to children with props.
I have added a test with 2 instance of vfg on the same page and they are isolated from each other.

lionel-bijaoui:
Groups are working. There is a lot of changes.
`form-group` is now `form-element` and `form-group` is a recursive component that allow for the flexibility enabled by this new system.
`form-element` should be seen as a wrapper of fields. It take care of label, help, buttons, hint and error display, as well as the field in itself. It also take care of `fieldID` and classes.

A group is a special type of field.
```js
{
    type: "group",
    legend: "My new group",
    styleClasses: "nice-group",
    tag: "section",
    fields: [...]
}
```
In a way, `schema` is just the main group of your form.

Now you can do that without problem (taken from the new grouping dev example)
```js
schema: {
    fields: [
        {
            type: "group",
            legend: "Contact Details",
            tag: "div",
            fields: [
                {
                    type: "input",
                    model: "name",
                    label: "Name",
                    fieldOptions: {
                        inputType: "text"
                    },
                    required: true,
                    validator: ["required"]
                },
                {
                    type: "group",
                    legend: "Subgroup",
                    styleClasses: "subgroup",
                    tag: "fieldset",
                    fields: [
                        {
                            type: "input",
                            model: "subname",
                            label: "Name",
                            fieldOptions: {
                                inputType: "text"
                            },
                            required: true,
                            validator: ["required"]
                        }
                    ]
                },
                {
                    type: "input",
                    model: "email",
                    label: "Email",
                    fieldOptions: {
                        inputType: "email"
                    }
                }
            ]
        },
        {
            type: "input",
            model: "single",
            label: "Single field (without group)",
            fieldOptions: {
                inputType: "text"
            },
            required: true,
            validator: ["string"]
        },
        {
            type: "group",
            legend: "Other Details",
            fields: [
                {
                    type: "input",
                    model: "others.more",
                    label: "More",
                    fieldOptions: {
                        inputType: "text"
                    }
                },
                {
                    type: "input",
                    model: "others.things",
                    label: "Things",
                    fieldOptions: {
                        inputType: "text"
                    }
                }
            ]
        }
    ]
}
```

Many things have been changed, I tried to stop using Lodash `forEach` and use ES6 `forEach` instead. Nothing against Lodash, but I think the native implementation is good enough.
Other time, I used Lodash `isNil`, `isFunction`, etc instead of manually checking.

I also changed the way the fieldUID are created. Before, I was using the internal `vm._uid` but now it use a mix of `fieldID` and Lodash `uniqueID`. For now this uid is use very loosely, but I think that having them could be useful for future needs.

Now for the caveats and limitations I was able to detect.
Group don't validate. Validation is using an event bus and is uncorrelated to structure. Events don't traverse the structure tree so group don't know the validation state of their children.
There is possible solution to find this information and maybe uid are also opening other way to do that. I will not fix it right now (will be explored in future patch/minor version).

Manual validation return a promise and is asynchronous. Since everything communication through the event bus (and by event), there is no way to fall back toward synchronous validation for the whole form.
It doesn't change the way single validator works as far as I know.
```html
<vue-form-generator ... ref="form"><vue-form-generator>
```

```js
/* Old way */
myManualValidation() {
    let errors = this.$refs.form.validate();
    if(errors.length > 0) {
        // Validation error
        console.warn("Error during validation", error);
    } else {
        // No validation errors
        // ...

    }
}
/* New way */
myManualValidation() {
    this.$refs.form.validate().then(
        () => {
            // No validation errors
            // ...
        },
        (error) => {
            // Validation error
            console.warn("Error during validation", error);
        }
    );
}
```

lionel-bijaoui:
I'm going to work on the validation states now.
The idea is to implement a `dirty` variable to have a third validation state (untouched, error and valid).
I will have a lot of work toward tests and documentation. While working on that I noticed a lot of missing documentation or errors in it. This update will be jarring to many users since it will require them to change a lot of options. We cannot release something without an good documentation.
@vue-generators/vfg In the meantime, please challenge and test my PR. I certainly have done stupid things and your feedback is priceless. Thank you all !

lionel-bijaoui:
I forgot to remove some big `skip`.
They are all about the new validation which is now a promise and not a synchronous validation.

Issue #244, opened 2017-06-30
Improve the field loading method


lionel-bijaoui:
> This is a following to the discussion started in #232 

The goal is to make field loading more modular, so that you only use the fields you really need.
@cristijora proposed this solution:
> adding component names exporting fields in a separate file

//input field
```
export default{
name:'vfgFieldInput'
}
```

//fields index file
```js
export default {Input, Checkbox} etc 
```

```js
install(Vue,options) {
     Vue.component("VueFormGenerator", module.exports.component);
     if(option.fields){
       fields.forEach((field)=>{
         Vue.component(field.name,field);
       })
     }
}
```

Usage:
```js
import VFG from 'vue-form-generator'
import {Input, Checkbox, AutoComplete} from 'vue-form-generator/fields'
Vue.use(VFG, [Input,Checkbox,Autocomplete])
```

This solution is a sweet spot between one big pack and only modular field. It allow to use internal component and custom ones, and take advantages of webpack tree-shaking feature.

Now, this also add some change to the plugin and some create some problem that we need to fix for v3.

Issue #32, opened 2016-08-10
create a full example on webpackbin


icebob:
Like the dev example with all supported fields.

lionel-bijaoui:
Every time I add `vue-form-generator`, it webpackbin fail... I give up :(

icebob:
I try it

icebob:
Same issue
https://github.com/christianalfoni/webpack-bin/issues/169

icebob:
webpackbin deprecated.

icebob:
webpackbin working (partially)
https://www.webpackbin.com/bins/-Ki03eOV65nKuBc_GLy4

icebob:
TODO: add link to README & Docs

lionel-bijaoui:
webpackbin is dead https://github.com/cerebral/webpackbin/issues/56

codesandbox.io is a replacement.
I'm going to create a repo with the dev example and import it.
Then I will add this as a submodule into the main repo.
That way exemple and dev will be the same.

PR #505, opened 2018-10-04
Improve field loading method

lionel-bijaoui:

Feature

#244 

There is a new way to deal with fields.
VFG used to register core and/or optionnal fields in Vue (depending on the version you decided to use).
That mean that internally, VFG was using `Vue.component()` to register all fields.
So if you wanted to use only one field or two, your final app had at best, all "core" fields bundled inside.

Now, no fields are added to Vue by default.
You need to pass them as an option.
```js
// Before
import VueFormGenerator from "vue-form-generator";
Vue.use(VueFormGenerator); // All fields are passed through Vue.component()

// Now
import VueFormGenerator from "vue-form-generator";
import {
  fieldCheckbox,
  fieldInput,
  fieldSubmit
} from "vue-form-generator/utils/fieldsLoader.js";
Vue.use(VueFormGenerator, {
  fields: [fieldCheckbox, fieldInput, fieldSubmit] // Only those fields will pass through Vue.component()
});
```
It still work with custom fields. Meaning now, you can use VFG with your custom fields only.
(__If you use a recent version of Webpack or a tool that does tree-shaking__, you should have a better bundle size.
Your custom field will need to have the ["name" property](https://vuejs.org/v2/api/#name) set to "field-***" (e.g. "field-awesome")
```js
import VueFormGenerator from "vue-form-generator";
import fieldAwesome from "./fieldAwesome.vue";
Vue.use(VueFormGenerator, {
  fields: [fieldCheckbox, fieldInput, fieldSubmit] // Only those fields will pass through Vue.component()
});
```
If you want everything like before (full bundle), you need to do that
```js
import VueFormGenerator from "vue-form-generator";
import {
  fieldCheckbox,
  fieldChecklist,
  fieldInput,
  fieldLabel,
  fieldRadios,
  fieldSelect,
  fieldSubmit,
  fieldTextArea,
  fieldUpload,
  fieldCleave,
  fieldDateTimePicker,
  fieldGoogleAddress,
  fieldImage,
  fieldMasked,
  fieldNoUiSlider,
  fieldPikaday,
  fieldRangeSlider,
  fieldSelectEx,
  fieldSpectrum,
  fieldStaticMap,
  fieldSwitch,
  fieldVueMultiSelect
} from "vue-form-generator/utils/fieldsLoader.js";

Vue.use(VueFormGenerator, {
  fields: [
    fieldCheckbox,
    fieldChecklist,
    fieldInput,
    fieldLabel,
    fieldRadios,
    fieldSelect,
    fieldSubmit,
    fieldTextArea,
    fieldUpload,
    fieldCleave,
    fieldDateTimePicker,
    fieldGoogleAddress,
    fieldImage,
    fieldMasked,
    fieldNoUiSlider,
    fieldPikaday,
    fieldRangeSlider,
    fieldSelectEx,
    fieldSpectrum,
    fieldStaticMap,
    fieldSwitch,
    fieldVueMultiSelect
  ]
});
```

* **Other information**:

Be aware that this is the first step toward removing fields from the library and into their own repository.
Soon, VFG will bundle no field at all, and you will need to add them as dependency.

Also, VFG is completely written in ES6 now (`import/export`), no more CommonJS syntax (`require/module.exports`). The mix of both caused a bug in Webpack.

### DIFF
--- a/dev/projects/basic/main.js
+++ b/dev/projects/basic/main.js
@@ -1,7 +1,10 @@
 import Vue from "vue";
 
 import VueFormGenerator from "@";
-Vue.use(VueFormGenerator);
+import { fieldCheckbox, fieldInput, fieldSubmit } from "@/utils/fieldsLoader.js";
+Vue.use(VueFormGenerator, {
+	fields: [fieldCheckbox, fieldInput, fieldSubmit]
+});
 
 import VueHighlightJS from "vue-highlightjs";
 Vue.use(VueHighlightJS);
--- a/dev/projects/checklist/app.vue
+++ b/dev/projects/checklist/app.vue
@@ -16,7 +16,8 @@
 
 <script>
 /* eslint no-console: 0 */
-import { validators } from "../../../src";
+import VueFormGenerator from "../../../src/index.js";
+const validators = VueFormGenerator.validators;
 import mixinUtils from "../../mixins/utils.js";
 
 export default {
--- a/dev/projects/checklist/main.js
+++ b/dev/projects/checklist/main.js
@@ -1,7 +1,10 @@
 import Vue from "vue";
 
 import VueFormGenerator from "@";
-Vue.use(VueFormGenerator);
+import { fieldChecklist } from "@/utils/fieldsLoader.js";
+Vue.use(VueFormGenerator, {
+	fields: [fieldChecklist]
+});
 
 import VueHighlightJS from "vue-highlightjs";
 Vue.use(VueHighlightJS);
--- a/dev/projects/custom/main.js
+++ b/dev/projects/custom/main.js
@@ -1,7 +1,10 @@
 import Vue from "vue";
 
 import VueFormGenerator from "@";
-Vue.use(VueFormGenerator);
+import { fieldInput } from "@/utils/fieldsLoader.js";
+Vue.use(VueFormGenerator, {
+	fields: [fieldInput]
+});
 
 import VueHighlightJS from "vue-highlightjs";
 Vue.use(VueHighlightJS);
--- a/dev/projects/full/app.vue
+++ b/dev/projects/full/app.vue
@@ -36,7 +36,7 @@
 <script>
 /* eslint no-console: 0 */
 import Vue from "vue";
-import VueFormGenerator from "../../../src";
+import VueFormGenerator from "@";
 import DataTable from "./dataTable.vue";
 import Fakerator from "fakerator";
 
@@ -47,13 +47,7 @@ import mixinUtils from "../../mixins/utils.js";
 import Multiselect from "vue-multiselect";
 Vue.component("multiselect", Multiselect);
 
-// Test custom field
-import FieldAwesome from "./fieldAwesome.vue";
-Vue.component("fieldAwesome", FieldAwesome);
-
 import { each, cloneDeep, merge } from "lodash";
-
-Vue.use(VueFormGenerator);
 
 let fakerator = new Fakerator();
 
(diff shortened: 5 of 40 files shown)


zoul0813:
is there a way to implement a quick and easy "import all" option?  This way we don't have to explicitly list the fields we would have access too.  Having to identify each field might be a bit problematic in apps that use dynamic forms ... where we don't know which fields a form will use ahead of time, and therefore have to have all fields loaded and ready to go?

`import { CoreFields } from '@vue-form-generator/fields';`

Where `CoreFields` or whatever it would be called, would export an array that can just be passed into Vue.use to register all "core" fields?

lionel-bijaoui:
> 
> 
> is there a way to implement a quick and easy "import all" option? This way we don't have to explicitly list the fields we would have access too. Having to identify each field might be a bit problematic in apps that use dynamic forms ... where we don't know which fields a form will use ahead of time, and therefore have to have all fields loaded and ready to go?
> 
> `import { CoreFields } from '@vue-form-generator/fields';`
> 
> Where `CoreFields` or whatever it would be called, would export an array that can just be passed into Vue.use to register all "core" fields?

Remember that we want to remove all fields from VFG. Soon that will not be possible anyway. You will have to add the fields in your dependencies.
That's a trade-off ? 🤷‍♂️

zoul0813:
I thought we were keeping the standard/core fields ... such as input, select, checkbox, etc ... as these are the backbone of all other fields and common across just about any form?

Not sure I agree with removing "all" fields from VFG Core.

lionel-bijaoui:
>
>Having to identify each field might be a bit problematic in apps that use dynamic forms

They can register all the fields if they want. In the end, this only use `Vue.component()`.

>
>I thought we were keeping the standard/core fields ... such as input, select, checkbox, etc ... as these are the backbone of all other fields and common across just about any form?
>
>Not sure I agree with removing "all" fields from VFG Core.

Don't worry, this is open for debate, but when we talked about this earlier, I understood you wanted to remove them all.
We could keep "core" field in VFG, move the rest into their own repo. No problem! 😄 

Tomorrow (or later), I will do a PR that remove Pug and use more strict EsLint rules.
Pug is stopping `eslint-plugin-vue` from doing it's job properly and I think this is an obstacle to new contributor. `eslint-plugin-vue` in vue cli 3 can fix almost everything in a simple `npm run lint`.

zoul0813:
Moving all the core fields into a single package, and flagging it as a dep of the main package is an ideal solution.  That would allow us to keep the main package smaller, and focused on the task of "generating the form" and the core-fields package could be focused on presenting the common "everyday" fields.  AbstractField and other common-reusable bits should still be part of the main form-generator package though, so third-party fields can extend from main... say for example, in forms where all elements are "custom" (for specific CSS frameworks, etc).

This would allow us to have a "core-bootstrap" and "core-blueprint" package for example, where the common fields are tailored toward either Bootstrap or Blueprint CSS frameworks, etc.

lionel-bijaoui:
I really want to do a repo by field, and avoid big pack like this.
If I update `field-input`, I don't want to have to update a big pack, just `field-input`.
If there is an issue, I want it focused on this single field.
If someone want to contribute to only this one field they use, they can.
We could use the embedded Wiki for documentation and make it super easy.

In this I see also a need to group `abstract field`, `slugify`, `dateFieldHelper` in a npm module (`vfg-utils` ?), so it could be used by fields without the need of all VFG.
Same thing with `validators`. Maybe use the same system (`options.validator`) to include other validation system ?

The "dev" folder need to be made into a [submodule](https://blog.github.com/2016-02-01-working-with-submodules/), so we can have a repo just to launch example (for #32 )

Issue #506, opened 2018-10-05
Submit button does not call form validator


qduc:
- **Do you want to request a _feature_ or report a _bug_?**
Report a bug

* **What is the current behavior?**
Given 'validateBeforeSubmit' is true and form data does not pass validation rules
When I click submit button, onSubmit function is called
- **Demo**
https://jsfiddle.net/nzkm7put/ (please check the result in console)

* **What is the expected behavior?**
onSubmit event should not be called

* **Please tell us about your environment:**

  * Version: 2.3.1
  * Browser: Chrome 69
  * Language: ES6/7

- **Other information:**

This may look like #434 but I'm using the built in onSubmit method, no form wrapper here so I think it is a different issue.

I have looked into the source code and found the call stack as below:

- fieldSubmit.onClick()
- formGroup.validate() <-- I think the problem is here, this should call the parent's validate() instead of child's one
- abstractField@validate() <-- submit button does not have any validator, of course this will pass, that's why there is no error

lionel-bijaoui:
Hello @qduc , you seem to know how to fix this, can you do a PR ?
We are all very busy, and that would help us tremendously. Thanks a lot !

qduc:
Sure, I am glad to help, just want to make sure it is a bug and not an intended behavior.
I will give it a try then.

lionel-bijaoui:
@qduc Oh, yes, I did not checked (sorry, I'm busy 😅 ).
In your fiddle, not error are triggered. I don't know why.
For the field, there is no problem (even though there should be one).
I'm sorry, I have no time to really look into it.
Can you investigate this absence of error ?

lionel-bijaoui:
I don't know, I fixed most of this in the v3, since the structure is no more about calling the parent, but sending event.

zoul0813:
@qduc this should be fixed in 2.3.2, can you confirm and close the issue if so?

dGilli:
@zoul0813 @qduc Can confirm just upgraded now and fixed for me. thx

qduc:
Yup, can confirm It's been fixed in 2.3.2.
Thanks for the release.

PR #509, opened 2018-10-08
Fix fieldSubmit not calling validate method

qduc:

Bug fix

#506

Beside the bug fix, _schema.onValidationError_ should be optional.
Previous behaviour, when validation failed, _onSubmit_ still get called if _onValidationError_ is not exists.

No

* **Other information**:

### DIFF
--- a/src/fields/core/fieldSubmit.vue
+++ b/src/fields/core/fieldSubmit.vue
@@ -4,7 +4,7 @@
 
 <script>
 import abstractField from "../abstractField";
-import { isFunction, isEmpty } from "lodash";
+import { get as objGet, isFunction, isEmpty } from "lodash";
 
 export default {
 	mixins: [abstractField],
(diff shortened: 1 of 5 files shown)


lionel-bijaoui:
@qduc Thanks a lot for this very good PR !

src/fields/core/fieldSubmit.vue
```diff
@@ -15,10 +15,13 @@ export default {
 				// prevent a <form /> from having it's submit event triggered
 				// when we have to validate data first
 				$event.preventDefault();
-				let errors = this.$parent.validate();
+				let validateAsync = objGet(this.formOptions, "validateAsync", false);
+				let errors = this.vfg.validate();
 				let handleErrors = errors => {
-					if (!isEmpty(errors) && isFunction(this.schema.onValidationError)) {
-						this.schema.onValidationError(this.model, this.schema, errors, $event);
+					if ((validateAsync && !isEmpty(errors)) || (!validateAsync && !errors)) {
```
zoul0813: Why the addition of this extra conditional? Haven’t had my coffee yet, but it seems to be redundant?

qduc: Because at this point, we still don't know if this form is valid or not.

If you mean there is already a check here
`if (errors && isFunction(errors.then))`

Because of the `isFunction(errors.then)` condition, a non-async valid result still go to the `else` clause and calls `handleErrors` function

qduc:
The reason I changed this logic is because I think it is unexpected to enable _validateBeforeSubmit_ but the form will call onSubmit anyway, whether it is valid or not.

So we let the user decide if they still want to submit or not but we don't provide a way for them to retrieve the validation result or an error array. As far as I can see, onSubmit's arguments doesn't have that data.

I'm sorry that I make this change without discussion first, if you think this is intentional and should be kept that way I'll make another PR to revert this.

lionel-bijaoui:
If this is really breaking thing, I will revert this commit.
I'm sorry for merging so fast, it looked good to me.

zoul0813:
I don’t think it’s “breaking”, but we need to clearly state the change in the release logs

lionel-bijaoui:
Ok, sorry again

PR #519, opened 2018-10-16
formGroup now use the correct class name for error when looking for them

lionel-bijaoui:

Bug fix

`form-group` listen for "field-validated" to trigger a search in their DOM childrens for element with a class of ".form-element.error".
If they find some, they apply an error class. If not, they add a validation class.
This is a very basic system to detect error in the children of groups (PR welcome if you have an idea to improve this without it being too costly).
The problem is that the error class can be customize with the option `validationErrorClass`.
The current `querySelector` doen't take this into account.
If you changed `validationErrorClass`, this basic verification system would break and show validation everytime, regardless of the content and state of the children of the group.

It now use `validationErrorClass` to get the error class name.

No

* **Other information**:

### DIFF
--- a/src/formGroup.vue
+++ b/src/formGroup.vue
@@ -139,7 +139,10 @@ export default {
 	created() {
 		this.eventBus.$on("field-validated", () => {
 			this.$nextTick(() => {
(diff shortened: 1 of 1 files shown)


zoul0813:
looks good to me

Issue #511, opened 2018-10-10
@2.3.1 input fields do not display text.


tizl1234:
Hi,
* **I'm submitting a ...**
  * [x] bug report
  * [ ] feature request
  * [ ] support request => Please do not submit support request here, see note at the top of this template.

* **What is the current behavior?**
VFG@2.3.1 Some input fields do not display text or any other data type ( first three in the example below), but the data changes in JSON.
- **If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem** via
  [codesandbox](https://codesandbox.io/s/m5zr4j88)
with pretty JSON output:
[fiddle](https://jsfiddle.net/tizl1234/8vwcy23b/3/)

* **What is the expected behavior?**
VFG@2.3.1 all input fieldes shows data.
- **What is the motivation / use case for changing the behavior?**
I suppose it's happening, because of the dot (.) or/and brackets ([]) operator in `model` property in ` schema` (e.g. `model:" mainBroker.url "`)
* **Please tell us about your environment:**

  * Version: [ 2.3.1]
  * Browser: [ Chrome 69.0.3497.100 ]
  * Language: [ES6/7]

I'm quite new in Vue.js and I hope I use VFG in proper way and I don't repeat already existed issue report.
You can change VFG version from 2.3.1 to 2.2.2. in codesandbox or jsfiddle and see the difference.

qkdreyer:
This looks like a major breaking change to me.

v2.3.1 : broken (https://jsfiddle.net/8cxh1nyg/1)
v2.3.0 : working (https://jsfiddle.net/8cxh1nyg/2)

qkdreyer:
@lionel-bijaoui Could this be caused by lodash-webpack-plugin ? (https://github.com/vue-generators/vue-form-generator/compare/v2.3.0...v2.3.1)

zoul0813:
should be fixed in 2.3.2

PR #526, opened 2018-10-17
Fix deep property path not working

lionel-bijaoui:

Bug fix

#511

No

* **Other information**:

- Deactivate `paths` option in 'lodash-webpack-plugin'

Thanks @qkdreyer


lionel-bijaoui:
@zoul0813 Yes !

zoul0813:
@lionel-bijaoui I just updated one of my projects to the 2.3.2 branch on my fork and I'm still not getting deep property paths loading up ... 

when I remove the LodashModuleReplacementPlugin from the webpack config, everything works fine ... when it's included, my "location.city, location.state, location.postalCode" fields don't display values in the inputs.

I think the "opt-in" config is backwards ... you pass "true" to enable the feature in the plugins options object.

I'm not sure which Lodash features we should keep enabled though ...paths and collections seems like something we should enable.

PR #541, opened 2018-10-28
Update README.md : Added reference to how to do manual validation of the form.

enyosolutions:

- [X ] The commit message follows our guidelines
- [ X] Docs have been added / updated (for bug fixes / features)

docs update

no breaking changes

* **Other information**:

### DIFF
--- a/README.md
+++ b/README.md
@@ -189,6 +189,27 @@ export default {
 };
 ```
 
+## Manual validation 
+
+```
+<vue-form-generator ... ref="form"><vue-form-generator>
+
+...
+
+myManualValidation() {
+    let errors = this.$refs.form.validate();
+    if(errors.length > 0) {
+        // Validation error
+        console.warn("Error during validation", error);
+    } else {
+        // No validation errors
+        // ...
+
+    }
+}
+
+```
+
 ## Development
 
 This command will start a `webpack-dev-server` with content of `dev` folder.


zoul0813:
This example isn’t accurate.  async validation returns a promise, and v3 will be moving toward this as a standard...

enyosolutions:
This example is accurate for the current version, it's probably best to add the two versions ?

zoul0813:
V2 supports async validation, when enabled in form options

enyosolutions:
@zoul0813 sorry to ask but which form options. 
Both the readme and the gitbook references only 3 options 


Property | Type | Description
-- | -- | --
validateAfterLoad | boolean | Execute validation after model loaded
validateAfterChanged | boolean | Execute validation after every change
fieldIdPrefix | string | Prefix to add to every field's id. Will be prepended to ids explicity set in the schema, as well as auto-generated ones.

zoul0813:
@enyosolutions the documentation is a bit behind, validateAsync was [added quite a while ago](https://github.com/vue-generators/vue-form-generator/commit/5a26ef1b4f4aaccd0c84137cd1edd7ecb0c8bce1) to support proper Async Validation.

It's also going to be the way V3 functions by default.

I'm closing this issue for now, as I don't think that manual validation samples should be in the README, and instead, should be updated in Gitbooks.

@lionel-bijaoui do we have access to the Gitbook Docs to update them? If not, should we switch over to something new so we can start properly updating the docs?

PR #549, opened 2018-11-15
Fix instantiation of datetimepicker

MikeMnD:
Well after a little testing and debugging this is what I found. 
datetimepicker should be called on the input element and till now it was called on the 
<div class="form-group valid col-md-4 col-lg-3 col-sm-6 field-dateTimePicker">
and it should be on:
let input = this.$el.querySelector(".form-control");

* **Other information**:


zoul0813:
this needs a bit more work.  We need to disable the datetimepicker `onDestroy`, and at the moment, that is being done against this.$el, and not `let input`.  We need to make `input` a data attribute, assign it's value, then call destroy on it later.

PR #560, opened 2018-12-10
added "options" to VFG install function, appending custom "validators…

### LINKED ISSUE #545: Global custom validators
* **I'm submitting a ...**
  * [ ] bug report
  * [x] feature request
  * [ ] support request

- **Do you want to request a _feature_ or report a _bug_?**

I want to request a feature.

* **What is the current behavior?**

Right now, there is a limited number of global field validators that come predefined with the plugin.

- **What is the motivation / use case for changing the behavior?**

Validators included by default do not cover a sufficient variety of possible validation criteria. For example, there is no validator that would require the user to 'check' a checkbox element (validating a value to be true). 

Currently, this functionality can be extended by passing custom functions to the validator field in the schema object, but this is a deviation from the API used for default validators (passing them as a string array). My feature proposal is to give developers the ability to pass their custom validators in the ```Vue.use(VueFormGenerator, ....)``` plugin inclusion call alongside custom fields. These validators could then be accessed in the schema under the ```validator``` array the same way predefined validators are.

zoul0813:
…" to the validators object that are passed into `Vue.use(VueFormGenerator, { validators: { key: (value, field, model) => {} })

Feature

VFG has a limited number of "built-in" validators, and custom validators require a reference to the function which is not possible when using JSON returned from the server.

Introduce the ability to modify the "built-in validators" when calling `Vue.use(VueFormGenerator)` by passing an options object with the validators.

No
* **Other information**:
Resolves #545 

Example usage:
```
import Vue from "vue";
import VueFormGenerator from "../../../src";
Vue.use(VueFormGenerator, {
	validators: {
		customValidator: (value, field, model) => {
			console.log("customValidator", value, field, model);
			return true;
		}
	}
});

import App from "./app.vue";

new Vue({
	...App
}).$mount("#app");
```

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -3,8 +3,13 @@ const schema = require("./utils/schema.js");
 const validators = require("./utils/validators.js").default;
 const fieldComponents = require("./utils/fieldsLoader").default;
 const abstractField = require("./fields/abstractField").default;
-const install = (Vue) => {
+const install = (Vue, options) => {
 	Vue.component("VueFormGenerator", module.exports.component);
+	if(options && options.validators) {
+		for(let key in options.validators) {
+			validators[key] = options.validators[key];
+		}
+	}
 };
 
 module.exports = {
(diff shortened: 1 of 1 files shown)


zoul0813:
docs added in https://github.com/vue-generators/vue-form-generator-docs/pull/4

Issue #393, opened 2018-02-04
VFG CSS overrides Bootstrap3.x classes, limiting form capabilities


TheStoof:
Hello,

For this issue, I refer to https://jsfiddle.net/1cmf4eur/1/ for the desired effect and to https://jsfiddle.net/awon7otp/1/ to observe the conflicts.

For those wishing to implement multi-column forms, you could eliminate vfg.css or vfg-core.css, but doing so means you'd forgo '.form-group.error' and '.form-group.error errors', 'span.help' and '.helper classes', '.form-group.required', and more. 

I feel that vfg.css/vfg-core.css implement Bootstrap-like styling, but ultimately is not as Bootstrap-friendly as it could be.

What do you think about further de-coupling the VFG-specific CSS styling (error, help, required, etc)?

zoul0813:
I think we should continue using the bootstrap compatible classes, but make the rest of the markup/classes more bootstrap compatible.  This will allow you to not use the VFG CSS and rely entirely on Bootstrap for styling, but if you'd prefer not to use bootstrap the VFG CSS can provide all the necessary default styling as it always has.  

The end result here is that you can eliminate the VFG CSS if your site already uses bootstrap and increase performance and reduce overhead by not loading "yet another stylesheet".

dflock:
Agree with @zoul0813. Am currently using vfg in a new project with Bootstrap 4 and this causes problems there too - you need to override a load of the CSS from vfg that's baked into `formGenerator.vue`, or take it out and use your custom version of vfg.

Would be ideal if the classes were totally configurable and there was no CSS baked into vfg, but in the meantime, just having no baked in CSS (or the very minimum) - and leave the clases alone, would be an improvement, I think.

icebob:
Can it solve the layout problem? https://adamwathan.me/renderless-components-in-vuejs/

k8n:
In the very least, perhaps, [formGroup.vue](https://github.com/vue-generators/vue-form-generator/blob/fa934e84e57d4ec1dc4c6c9889aa2b1c279fe2b9/src/formGroup.vue#L123) could wrap it's SCSS in `.vue-form-generator { }` just like [formGenerator.vue](https://github.com/vue-generators/vue-form-generator/blob/fa934e84e57d4ec1dc4c6c9889aa2b1c279fe2b9/src/formGenerator.vue#L211) does?

PR #577, opened 2019-01-10
form-group css class specific to vue-form-generator

k8n:
... to avoid clashing with Bootstrap and others (#393)

bug fix

form-group css class definition is greedy, clobbers Bootstrap's

form-group class definition on applies to children of vue-form-generator class

Hopefully not

* **Other information**:

### DIFF
--- a/src/formGroup.vue
+++ b/src/formGroup.vue
@@ -120,66 +120,68 @@ export default {
 </script>
 <style lang="scss">
 $errorColor: #f00;
-.form-group:not([class*=" col-"]) {
-	width: 100%;
-}
-.form-group {
-	display: inline-block;
-	vertical-align: top;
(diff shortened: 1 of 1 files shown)


zoul0813:
this should be a v3 fix, as it would introduce breaking changes to anything that uses VFG and may happen to inherit styles from VFG (intentionally or not) on the rest of the page.

I believe the best approach here is to provide a standard VFG CSS file, that isn't baked into the Vue components ... and then used [Scoped CSS](https://vue-loader.vuejs.org/guide/scoped-css.html) blocks within the components for any specific overrides or component only modifications.

@lionel-bijaoui - thoughts?

ScopZy:
Not currently fixed ?

PR #600, opened 2019-03-07
 Add the distinction between actual form errors and errors displayed

Rydra:
If any of its fields of a form are invalid the entire form should be considered invalid even if the errors are not displayed for example due to the `isNewModel` property (which does not show all field errors at the beginning). So I created the distinction between the actual errors of the form, and the errors that are being displayed in the form.

I was facing a use case where I needed to have a submit button disabled until the form was correctly filled, but neither show the validation errors until the user reached the field for the first time. With how isNewModel was implemented this behaviour was not possible, the errors were being cleaned but when `validated` event was raised the isValid value was always true, so I introduced this functionality. If you supply isNewModel to false (the default) the behaviour is the same as it was before.


zoul0813:
interesting... I'll try to review this more thoroughly and get the PR merged or my feedback back to you ASAP

PR #603, opened 2019-03-26
Updated bundle sizes

okydk:
The current sizes are very misleading - the core is almost 4 times bigger than advertised.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@ A schema-based form generator component for Vue.js.
 *   multiple object editing
 *   21 field types
 *   built-in validators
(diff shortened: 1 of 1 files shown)


zoul0813:
Any chance you can also submit these changes to the doc repo for gitbooks?

okydk:
I don't think bundle sizes are mentioned in gitbooks?

zoul0813:
The README for gitbooks should be a duplicate copy of the projects main README.  I may have removed the bundle sizes from it recently for the same reason you updated them here.

If it’s not there, no big deal...

PR #606, opened 2019-04-08
Group visibility control

k8n:
Simple, but perhaps naïve change to permit `visible` control for groups same as for fields.

Ability to control visibility of field groups.

Groups visibility cannot be controlled. Groups are always rendered.

Specifying `visible` attribute on a field group permits to control group visibility.

Don't think so.


zoul0813:
Looks good, simple and gets the job done ... get another contributor to approve and I’ll merge it

muschottkey:
👍 

I'm waiting on this feature to be added so that I can use the module.

Can we get it approved please? 😀

PR #618, opened 2019-06-20
Fix fieldComponents exposure

muschottkey:
Exposing fieldComponents was broken because we were exporting named
exports in the `fieldsLoader.js` while we were importing as though it was exported as a default export.

Bug fix

In export (`utils/fieldsLoader.js`):

```javascript
module.exports = fieldComponents;
```

In import time inside (`index.js`)

```js
const fieldComponents = require("./utils/fieldsLoader").default;
```

This meant that this code would not be doing the right thing.

```js
import { fieldComponents } from 'vue-form-generator'
```
Because 

```js
// fieldComponents === undefined; because we imported it as if it was a default export
```

I'm sure this is not intended.

`fieldComponents` would always be `undefined`, which isn't very nice because for some
usecases, you might actually want to have access to the components.

Changes the importing to 
```js
const fieldComponents = require("./utils/fieldsLoader");
```

No breaking changes introduced.

* **Other information**:

Nothing else to add.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -1,7 +1,7 @@
 const component = require("./formGenerator.vue").default;
 const schema = require("./utils/schema.js");
 const validators = require("./utils/validators.js").default;
-const fieldComponents = require("./utils/fieldsLoader").default;
+const fieldComponents = require("./utils/fieldsLoader");
 const abstractField = require("./fields/abstractField").default;
 const install = (Vue, options) => {
 	Vue.component("VueFormGenerator", module.exports.component);


hunterae:
I am also experiencing this as an issue - namely, i am building another component that wraps around FieldCleave, but I have no way of importing FieldCleave directly except perhaps by adding all the necessary dependencies that VueFormGenerator uses internally. Ideally, I'd like to be able to grab the built definition off of fieldComponents but as mentioned above, fieldComponents is undefined because it is imported / exported incorrectly.

PR #639, opened 2019-08-09
Make is possible to hide/show a group

kevinkiel:


kevinkiel:
I just see its a duplicate https://github.com/vue-generators/vue-form-generator/pull/606

PR #651, opened 2019-10-18
Add better support for bootstrap horizontal form

muschottkey:
I wrapped the form element, along with its hint and error
messages with a div.field-content.

This provides separation of the label of the element from the
actual input element, which is necessary in order to have a proper
horizontal form design.

In bootstrap v4 horizontal forms, we have the label on the left hand side,
and the input element (along with accompanying content like hints
below it and validation messages) floated to the right.

Without wrapping the field in a div, the content under the input
would not be properly aligned with the input itself if the input
is offset from the left-most border of it's parent. This means
that horizontal forms wouldn't work well.

I decided to add another div instead of moving the hint and error
messages into the div.field-wrap because I think there might still
be use-cases where you would want to have a separate parent for the
field-element.

The inserted div shouldn't affect styling, since in css it can be defined
as block, flex, inline, inline-block e.t.c to the users needs.

* **Other information**:

### DIFF
--- a/src/formElement.vue
+++ b/src/formElement.vue
@@ -16,43 +16,45 @@
 				:getValueFromOption="getValueFromOption"></slot>
 		</label>
 
-		<div class="field-wrap">
-			<component
-				ref="child"
-				:is="fieldType"
-				:model="model"
-				:schema="field"
-				:form-options="options"
-				:event-bus="eventBus"
-				:field-id="fieldID"
-				@field-touched="onFieldTouched"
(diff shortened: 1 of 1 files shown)


RGKrish183:
Is it possible to achive like below pics, i have tried many way but didnt get any solutions, pls help to completed this. 

Under input box i have place **hint**, if valid it should always grey, if there have any errors need change it as red.

image

Thanks,
Gopal R

PR #667, opened 2020-03-03
Add form field align.

Wesamgerges:

* **Other information**:


DelfsEngineering:
Could this not be better handled already  With `fieldStyleClasses` ?

Wesamgerges:
This is mainly for fields that require the text to me on its right like checkbox or radio.
I have tried to use css, however, I couldn't get it right.
You may have more experience in css than me.
If you have a better solution, please let me know :) 
Thanks

DelfsEngineering:
`labelClasses` and  `fieldClasses` classes are available. Perhaps they need to be implemented better or additional hooks for clases to be added?

zoul0813:
As @DelfsEngineering stated, you should use CSS for this.

If you find there is t enough support with the existing fieldClasses and labelClasses, please provide a JSFiddle example of your use case so we can take a look.

Wesamgerges:
iagree

I just need to do this.
I don't know if this can be done with just css.

zoul0813:
Use the required attribute and :after pseudo class?

Issue #285, opened 2017-09-09
Render groups as bootstrap tabs


ericpyle:
Wishing I could use groups to render the form in a different way, like with each section having it's own tab. I could have multiple vfg instances with one form per tab. Or maybe I could use multi attribute to just hide all fields that are not in that tab. 

Thoughts on best way to do this?

ericpyle:
Or use visibility function that can somehow determine which tab is in use?

DelfsEngineering:
This works great:
Take your source schema and insert a function for the `visibility` field key's value

So if you want to show/hide field(s) based on the value of another field set their visibility to a function like 

```
function(){
   return this.model.fieldA == 'someValue'
}
```

of course use what ever logic you want now.

DelfsEngineering:
For tabs see `vue-form-wizard` here https://github.com/cristijora/vue-form-wizard

ericpyle:
@DelfsEngineering thanks so much for the suggestions! I'll definitely take a harder look at the vue-form-wizard.  Know of any examples of how to make the tabs look like regular bootstrap tabs?  It's possible that I could rethink whether my app should have more of a guided wizard UX, but converting my app over to how it currently works would be great.

ericpyle:
@DelfsEngineering also, do you know of any examples of how the multi field and multiple option is used in practice? I see what the documentation says, but I suspect there's a use-case or use-cases behind supporting that and the merge logic that respects multi.  Thanks again for any help!

DelfsEngineering:
I suppose the best you could do without forking `vue-form-wizard` is some injected CSS, or hacking the accompanying CSS.

cristijora:
Hey @ericpyle 
For tabs 
https://github.com/spatie/vue-tabs-component
https://github.com/cristijora/vue-tabs

For a more strict flow/wizard
https://github.com/cristijora/vue-form-wizard
The wizard supports tab layout, however they do not act like tabs since the wizard has a stricter validation flow (e.g you can't click a tab unless it was visited and validated)
https://jsfiddle.net/bt5dhqtf/99/

If you need tabs use tab components, if you need a strict validation flow use the wizard. Both tabs and the wizard should work fine with form-generator.

ericpyle:
Thanks, @cristijora. I'll look into integrating vfg with the other tab components and also ask my team whether we want to move toward a wizard approach.

[...]

PR #670, opened 2020-04-18
Add subgroups + custom group tags

johnsusek:
* **What kind of change does this PR introduce?** 
  - A feature to add a subgroup within a group
  - A feature to specify group tag per group

No, feature addition only.

* **Other information**:
Please see https://github.com/vue-generators/vue-form-generator-docs/pull/28 for documentation

---

I've also added subgroups to the dev group example:

Screen Shot 2020-04-18 at 2 16 29 PM

---

Since these changes allow custom tags and subgroups, complex behavior like putting subgroups into tabs is possible. Here's what it looks like with bootstrap-vue tabs for subgroups:

Screen Shot 2020-04-18 at 2 17 12 PM

---

Please let me know if there's anything you would like done differently! I hope we can get this subgrouping behavior in, we are using vfg in a project with very complex forms and could really use something like this. 

Related issue: #285

### DIFF
--- a/dev/projects/grouping/app.vue
+++ b/dev/projects/grouping/app.vue
@@ -25,6 +25,9 @@ export default {
 			model: {
 				name: "Brian Blessed",
 				email: "brian@hawkman.mongo",
+				city: "Springfield",
+				state: "IL",
+				age: 25,
 				others: {
 					more: "More",
 					things: "Things"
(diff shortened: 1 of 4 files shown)


DelfsEngineering:
This is interesting but I think a `visible` key really needs to be part of this to make it really useful. I think a `visible` tag in on another PR.
I don't see a sample of what the schema would look like, I am assuming it can go multiple levels deep?

johnsusek:
For our use case, this functionality is definitely useful and would go into use right away to group forms into subsections. And other people would like it too imho, as evidence I would point to #285.

But maybe I don't understand - can you elaborate on this "visible" feature and why it is important for this PR?

Thanks for the feedback!

DelfsEngineering:
A `visible` key at the group level to hide the entire group, currently it must be done via css.

johnsusek:
That sounds like it would be a useful addition to vfg. If that feature is added to the codebase I would be happy to help integrate it with this subgroups feature.

DelfsEngineering:
Basically it would allow you to have a login, password reset and sign up groups all in one form.

DelfsEngineering:
@johnsusek Would this support recursive groups also? or just one level down?

Mohammad-Alavi:
This is a nice feature, why it is not merged yet? 🤔

Mohammad-Alavi:
This is a nice feature. Why it is not merged yet? 🤔

vortechron:
@icebob merge it please :(

PR #671, opened 2020-04-27
fix(validation): URL validation for URL(s) with more than four characters TLD

ismail9k:
This __PR__ is mainly to fix URL validation for URL(s) with more than four characters TLD (e.g. `.store`, `.online`, `.studio`, `.design`, `.international` ...)

Bug fix

It gives invalid URL message for URLs five-plus characters TLD (example this website [`https://mycatalog.store/`](https://mycatalog.store/))

URLs with more than four characters TLD should tread as a valid URLs.

No breaking change

* **Other information**:

### DIFF
--- a/src/utils/validators.js
+++ b/src/utils/validators.js
@@ -198,7 +198,7 @@ const validators = {
 		let res = checkEmpty(value, field.required, messages);
 		if (res != null) return res;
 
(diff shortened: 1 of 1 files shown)


zoul0813:
If we’re going to fix this, should address the fact that TLD’s do not have a length limit... to my knowledge.  There are some fairly large ones.  I believe “business” is a TLD now for example.

The TLD should be validated if it contains valid chars, regardless of length...

ismail9k:
Yes, we should support all the currently available TLDs. However, I have searched for the list of available TLD, the characters count vary from 2 to around 24 characters. So, I suggest validating TLD to be a minimum of 2 characters without maximum.

zoul0813 (approved):

ismail9k:
Any update on this  :), when this *PR* will be available on the release version?

